PR #13, opened 2019-05-07
refactor Makefile to avoid rebuilds

peterbe:
Now you can run `make lint` (or `make validate`) repeatedly without having to trigger the docker build every time.


peterbe:
@escattone By the way, I like this pattern. We should perhaps put something like it in our Makefile for kuma to avoid certain `make ...` triggering things that don't need to re-run. E.g. in Travis.

PR #28, opened 2019-05-22
prettier'ify all script code

peterbe:


peterbe:
I copied a bunch of these techniques from the good folks at https://github.com/mozilla/addons-frontend

wbamberg:
I'm very happy for us to use prettier, but is there a reason we using double quotes for strings? This is contrary to our practice on any other MDN project that I know of.

peterbe:
> is there a reason we using double quotes for strings? 

Actually, there's not even an [ADR](https://github.com/mdn/mdn/tree/master/ADRs) for the 4-space indentation. I just used that because I saw that in other .js code kuma.

Personally, I'm not fussed because the beauty of all of this is that it doesn't matter how you type it. With good tooling (e.g. prettier integrated in VSCode) you just hit Ctrl-S and the document is formatted. 

Ideally I'd like to use no configurations and leave it entirely to the defaults from Prettier but I can also change it to match whatever was the legacy.

peterbe:
I see that bcd is using 2-spaces and single quotes
https://github.com/mdn/browser-compat-data/blob/master/index.js

peterbe:
Actually, there's a bunch of double quotes in bcd too. 

David's react code seems to be 4-spaces and single quotes https://github.com/mozilla/kuma/blob/master/kuma/javascript/src/header/search.jsx

Elchi3:
This is a discussion I was hoping to have with Daniel and everyone at the all hands. Multiple projects want to use prettier, but we're unsure how to use it exactly. See also https://github.com/mdn/mdn/issues/71 and https://github.com/mdn/browser-compat-data/pull/3737.

peterbe:
I don't disagree nor do I have super strong opinions (*) here. 

And I would like to get the right integration combo of eslint and prettier but perhaps it can wait till the next cycle and just focus on Prettier here and now. 

(*) I do prefer 2-space indentation in general because almost every other code base I've ever worked on in the last couple of years have this. But for double vs single quotes, I genuinely don't care.

karlhorky:
> I'm very happy for us to use prettier, but is there a reason we using double quotes for strings? This is contrary to our practice on any other MDN project that I know of.

This is a common change to the default config.

There is a big discussion around changing the default to single quotes (albeit with no resolution yet).

https://github.com/prettier/prettier/issues/4102

But with some positive signals from the maintainers and creator ("we always meant to do that"):

https://github.com/prettier/prettier/issues/3503#issuecomment-352802184

peterbe:
Thanks @karlhorky for those pointers. 
That kind of feedback would actually be great to add to https://github.com/mdn/mdn/pull/73

That standard would affect stumptown-content too eventually. 
I think it's definitely value in messing with the config a little bit if necessary. For example, it might be worth for an existing project to adopt some overrides to avoid a crazy reformatting of way too much legacy code. 

Still though, the most important point is that it doesn't matter :)

peterbe:
Note to self, when https://github.com/mdn/mdn/pull/73 lands I'm going to "redo" this PR and change the config to be an empty config.

peterbe:
Note-to-self; the `huskey` script should run `pretty-quick` and the `js-lint` one too.

peterbe:
This'll be too hard to conflict resolve. I'll start from scratch.

karlhorky:
Looks like this continues in #187, for anyone else following along.

PR #29, opened 2019-05-22
MPL 2.0

peterbe:


ddbeck:
Just a heads up: I expect that, like https://github.com/mdn/short-descriptions/pull/1, licensing will be a little awkward for Stumptown. We'll probably want the CC-BY-SA license for the content and the published package and a more permissive license for other source (also, BCD and short-descriptions went with CC0—I'm certain CC0 is compatible with CC-BY-SA, but I don't know whether the same works for MPL and CC-BY-SA).

wbamberg:
Yes, Daniel is right. We'll need to make ./content CC-BY-SA.

As for the source, I think it would be good to be consistent with (most of) the other repos in [mdn](https://github.com/mdn/), which use CC0.

We should solicit opinions from the owners of mdn/mdn though, although I'm not sure who they are. @chrisdavidmills , @atopal , @Elchi3 , maybe?

chrisdavidmills:
Yes, this sounds correct to me, and is what we've been aiming for longer term Content licensed with the same license as the Wiki, data and code CC0.

There is no problem having licenses mixed in the same repo.

And the two are compatible with each other. I found this useful chart: https://en.wikipedia.org/wiki/License_compatibility#Creative_Commons_license_compatibility

Elchi3:
I agree with Daniel.

atopal:
Makes sense to me :+1:

peterbe:
The `package.json` is a part of the scripts. I.e. the executable Node things. Does that need to be different? 

Definitely feels like a slam dunk that the `LICENSE.txt` that this PR adds is wrong in that it's the LICENSE file for MPL-2.0.

ddbeck:
For short descriptions, we had two license files, one in the root directory (with a note at the top mentioning the other) and one in a content directory (see [this bit of the readme](https://github.com/mdn/short-descriptions#license)). The `package.json` file reports the more restrictive of the two licenses.

PR #33, opened 2019-05-28
Add partial HTML rendering

wbamberg:
For this sprint I'm working on some bits of https://github.com/mdn/sprints/issues/1638: writing some code to render MDN pages out of stumptown JSON.

I don't think the rendering code should live in this repo actually, so I really just filed this PR as a way to make it possible for people to try out the code and think about where we could house it best.

I did look at integrating it into https://github.com/peterbe/mdn2, but (a) that looked pretty complicated to me and (b) it looks like it needs some polish before being very contributor-friendly. But I'd be very happy to have a go at rehousing this code there if @peterbe can help me understand where to put it.

The only thing I really care about is for it to be modular, so it's easy to write code to render different bits of the page (e.g. compat tables, examples, etc) and drop them in.

In this version I'm basically following the pattern of build-json, where there's a single entry point that assembles a document out of pieces, and separate modules for each nontrivial piece.

So to run it:

* from a fresh checkout, first build the json: `npm run build-json html`
* then render it to static HTML: `npm run render-html video`

...it writes the file to ./rendered/html/elements/video.html and you can open it in a browser and it looks a bit like an MDN page with no CSS.

At the moment it includes:

* some prose sections (short descriptions, overview)
* interactive example
* attributes list (but not including possible values yet)
* live samples (using `srcdoc`, which I think will not work for CSS, but perhaps `src="data:..."` will?)

One tiny thing that makes me very happy: in the [MDN video page, the attributes list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#Attributes) is not properly ordered. Bad news if you are looking for `preload`! In this version, not only is it properly ordered, but it will *always* be properly ordered, without anyone having to think about it. For me this kind of thing is a big part of the motivation for this project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 node_modules/
 .docker-build
 packaged
+rendered
--- a/package.json
+++ b/package.json
@@ -17,9 +17,11 @@
         "prettify": "prettier --write scripts/**/*.js",
         "lint-md": "remark content",
         "build-json": "node scripts/build-json/build-json.js",
+        "render-html": "node scripts/render-html/render-html.js",
         "spell-md": "mdspell -a -n --en-us 'content/**/*.md'"
     },
     "dependencies": {
+        "escape-html": "1.0.3",
         "gray-matter": "4.0.2",
         "js-yaml": "3.13.1",
         "jsdom": "^12.2.0",
--- a/scripts/build-json/compose-attributes.js
+++ b/scripts/build-json/compose-attributes.js
@@ -9,10 +9,10 @@ const bcd = require('mdn-browser-compat-data');
 
 const { JSDOM } = jsdom;
 
-function extractFromSiblings(node, terminatorTag, contentType) {
+function extractFromSiblings(node, terminatorTags, contentType) {
     let content = '';
     let sib = node.nextSibling;
-    while (sib && sib.nodeName != terminatorTag) {
+    while (sib && !terminatorTags.includes(sib.nodeName)) {
         if (sib.outerHTML) {
             if (contentType === 'html') {
                 content += sib.outerHTML;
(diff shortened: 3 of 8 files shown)


peterbe:
I don't know where to start but can we please halt this effort? 
I already have a prototype that is trying to solve the exact same thing. Granted, my prototype is very "MDN heavy" but it already takes a good chunk out of https://github.com/mdn/sprints/issues/1638

Beyond the terrible name (*), https://github.com/peterbe/mdn2 is a good intention. It's going to be possible to depend on it within stumptown and use its cli to render out every single page, and create a netlify subdomain, with all rendered pages, for every PR on stumptown-experiment. 

The major difference is that https://github.com/peterbe/mdn2 is React based, works as a website (for viewing in a browser with PJAX support), and a cli that can convert any stumptown path into a viewable .html page. 

(*) `stumptown-experiment` isn't a great name either at this point since we're starting to take this experiment seriously. :)

peterbe:
I haven't been very verbose about https://github.com/peterbe/mdn2 because it's such early days and I totally agree that it's pretty complicated at the moment. We should join forces and not make it a "peter project". 
My next intention was to hack on the [`BrowserCompatibility` React component](https://github.com/peterbe/mdn2/blob/a5b9805626dbfc43dd28537f9d907952a88cfb44/client/src/App.js#L243) which is poised to produce a fully working blob of HTML/DOM based on the `__compat` data. I've only created a stub but the next task is to study the Compat.ejs code more in detail and slowly rewrite it as a React component. 

There are other stubs within that same `App.js` for rendering sections but I admit that it's not clear just yet how to dig in and share the code effort.

wbamberg:
As I said in the original PR, I'm very happy to rehouse this code or something functionally similar in mdn2. But for that to be possible mdn2 needs to be more contributor-friendly than it is today.

peterbe:
> But for that to be possible mdn2 needs to be more contributor-friendly than it is today.

I think you're more than right. If this is going to get some wings, I need to pause all the other hacks and make the contributor-friendly story better. 
One shouldn't need to read a bunch of complicated commands. Watch this space!

peterbe:
To hopefully close the loop a little bit here, the path for contributing on mdn2 is now improved; https://github.com/peterbe/mdn2#contributing

wbamberg:
Closed in favour of https://github.com/peterbe/mdn2/pull/3.

PR #36, opened 2019-05-29
Fix #16: Remove Python code

### LINKED ISSUE #16: Remove the Python code?
In https://github.com/mdn/stumptown-experiment/pull/14 we're developing the first cli tool that can be used to turn the .md and .yaml files into .json files. 

The next natural thing to do, especially from a kuma perspective, is to now glob for all these .json files and one by one, do a lookup of its recipe and then do something with that. E.g. extract the bits it needs to turn it into kuma content.

So what else do we need Python for. Today, the only entrypoint I see is the `validate.py` script.

Regarding https://github.com/mdn/stumptown-experiment/pull/14 @wbamberg mentioned that that script assumes that the integrity of all the core content (e.g. .md and .yaml files) is perfect. And I think that's right. 

I would be open to rewriting `validate.py` to `validate.js` and reduce this whole project to become a 100% Node project.

Elchi3:
I updated my local copy of stumptown and wanted to take a look around as I haven't looked at it in quite a while. I wanted to see if I can contribute content, but content validation seems to be somewhat stuck. (please tell if it is not and how to do it).

So, in this PR I removed the Python code as discussed in #16 and added travis runs for "lint-md", so that some initial validation happens.

I tried adding `spell-md` and `pretty` to travis runs, too, but those seem to fail currently, so I saved this for follow-ups. Also, there was some validation in validate.py, but that's also failing, so I thought that we should probably just get rid of it now and discuss how we want to validate what and then get that implemented as a follow-up as well.

A sample travis run can be seen here: https://travis-ci.com/Elchi3/stumptown-experiment/builds/113606455

If this doesn't suit into any current plans, feel free to throw this PR away.


wbamberg:
@Elchi3 , thanks for this and thanks @peterbe for the review.

We're more focused on rendering a single page in this sprint than validation, but validation is super-important of course, especially once we start trying to add lots of content.

The specifics of how things should be formatted are also still changing of course (see https://github.com/mdn/stumptown-experiment/issues/19 for a major one), but mostly it's solid enough for validation to be useful I think. Of course we can expect that to go on changing for a while, and we'll just have to deal with it.

The old Python code did some generic validation (for example spell check IIRC) and some very basic top-level validation. But things get more interesting in the details: like how exactly an example should be formatted? At that level validation is blocked on having good documentation about how things should be formatted.

PR #39, opened 2019-05-30
Re-enable spell checks for md files

Elchi3:
This PR re-enables spell-checking for markdown files and in my opinion it did uncover real spelling mistakes and inconsistencies in the md files. I'm a fan of this validation, but again, if this doesn't suit any current plans, ignore this PR for now or close it.

Summary of changes:

- Add back `mdspell` to the `test` command, so that the spellcheck runs when running `npm t` locally and when travis checks stumptown (I will enable travis soon hopefully).
- Add "!(*contributors).md" to mdspell glob to avoid looking into contributor md files at all.
- Remove a bunch of technical terms from .spelling, as we should mark them up in `<code>` (or rather using backticks in md)
- Consistently mark up HTML attributes h1s and the attribute values h3s in `<code>`. This avoids also things like ["\\_parent"](https://github.com/mdn/stumptown-experiment/pull/39/files#diff-df6f8e9c487507936fafc3d483a96667L30) and looks much better as `_parent`. Mention this convention in the docs. Follow-up could be to lint this so it is actually guaranteed.
- Avoid line breaks in links as the spell checker can't understand markdown links then and so throws many errors on URLs :/ (do we have a line length limit or why were these split? It also breaks my md syntax highlighter in VS code, fwiw)


Elchi3:
Travis runs now! See https://travis-ci.org/mdn/stumptown-experiment/builds/539196091?utm_source=github_status&utm_medium=notification and at the end it says:

>> 261 files are free from spelling errors

peterbe:
I might have to sit this one out, but I noticed that it's mostly changes from 
```
# cellspacing
```
to 
```
# `cellspacing`
```
which seems very right. Go for it!

Elchi3:
> This is just the output of Pandoc :(. Perhaps we could tell it not to do this.

That makes sense. Thanks Will!

PR #40, opened 2019-05-30
Fix bug in attribute value parsing

wbamberg:
The code in [compose-attributes](https://github.com/mdn/stumptown-experiment/blob/master/scripts/build-json/compose-attributes.js) is supposed to parse HTML attributes from the Markdown source into a JSON structure.

Among other things it tries to parse out attribute values, which are separated by H3 headings in the source. To do this it starts at the start and collects markup until it finds a tag that tells it to stop. In the old code this finishing tag was "H3", which works fine until the value is the last one, because then the next heading tag is the H2 for the attribute type, which then mistakenly gets swept up into the value description.

This change simply lets the client pass an array of tags, and we stop when any of them are encountered.

I've verified that with this fix, mdn2 renders attributes properly.

This PR also removes an unused copy of the function from slice-prose.

### DIFF
--- a/scripts/build-json/compose-attributes.js
+++ b/scripts/build-json/compose-attributes.js
@@ -9,10 +9,10 @@ const bcd = require('mdn-browser-compat-data');
 
 const { JSDOM } = jsdom;
 
-function extractFromSiblings(node, terminatorTag, contentType) {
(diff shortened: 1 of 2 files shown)


peterbe:
I've upgraded mdn2 to include this changeset now.

PR #41, opened 2019-05-30
Use headings rather than comments to demarcate prose sections

wbamberg:
This implements a new way of demarcating sections in prose.md, along the lines of https://github.com/mdn/stumptown-experiment/issues/19.

* update video's prose.md to use just H2 headings to mark top-level sections
* update the JSON builder code to understand that and to emit title and content separately, like:

```
"short_description": {
    "title": "Short description",
    "content": "<p>The <strong>HTML Video element</strong> (<strong><code>&lt;video&gt;</code></strong>) embeds a media player which\nsupports video playback into the document.</p>"
},
```

Because this changes the JSON it'll break mdn2: I'll file a separate PR to fix that.

It doesn't yet implement "registration" for prose.* sections, but anyway video doesn't have any so we don't need it yet.

It also updates the recipe to use only "_" as a space-replacement.

### DIFF
--- a/scripts/build-json/slice-prose.js
+++ b/scripts/build-json/slice-prose.js
@@ -10,13 +10,17 @@ const bcd = require('mdn-browser-compat-data');
 const { JSDOM } = jsdom;
 const commentNode = 8;
 const namedSections = [
-  'short-description',
-  'overview',
-  'attributes-text',
-  'usage-notes',
-  'accessibility-concerns',
-  'see-also'
+  'Short description',
+  'Overview',
+  'Attributes',
+  'Usage notes',
+  'Accessibility concerns',
+  'See also'
 ];
+
+function sectionNameToSlug(sectionName) {
+    return sectionName.toLowerCase().replace(' ', '_');
+}
 
 function extractFromSiblings(node, terminatorTag, contentType) {
     let content = '';
@@ -49,14 +53,18 @@ function packageValues(heading, dom) {
 
 function getSection(node, sections) {
     const sectionName = node.textContent.trim();
-    const sectionContent = extractFromSiblings(node, '#comment', 'html');
+    const sectionContent = extractFromSiblings(node, 'H2', 'html');
     const extraSections = [];
 
     if (namedSections.includes(sectionName)) {
-        sections[sectionName] = sectionContent;
+        const sectionSlug = sectionNameToSlug(sectionName);
+        sections[sectionSlug] = {
+            title: sectionName,
+            content: sectionContent
+        }
     } else {
         const additionalSection = {
-            name: sectionName,
+            title: sectionName,
             content: sectionContent
         };
       sections['additional-sections'].push(additionalSection);
@@ -71,7 +79,7 @@ function package(prosePath) {
     };
     let node = dom.firstChild;
     while (node) {
-        if (node.nodeType === commentNode) {
+        if (node.nodeName === 'H2') {
             getSection(node, sections);
         }
         node = node.nextSibling;
--- a/content/html/elements/video/prose.md
+++ b/content/html/elements/video/prose.md
@@ -1,7 +1,10 @@
-<!-- short-description -->
+## Short description
+
 The **HTML Video element** (**`<video>`**) embeds a media player which
 supports video playback into the document.
-<!-- overview -->
+
+## Overview
+
 You can use `<video>` for audio content as well, but the [`<audio>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio)
 element may provide a more appropriate user experience.
 
@@ -51,9 +54,8 @@ A good general source of information on using HTML `<video>` is the
 [Video and audio
 content](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Video_and_audio_content)
 beginner's tutorial.
-<!-- usage-notes -->
-Usage notes
------------
+
+## Usage notes
 
 ### Styling with CSS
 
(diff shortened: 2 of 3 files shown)


scripts/build-json/slice-prose.js
```diff
@@ -49,14 +53,18 @@ function packageValues(heading, dom) {
...
+    const sectionContent = extractFromSiblings(node, 'H2', 'html');
     const extraSections = [];
 
     if (namedSections.includes(sectionName)) {
-        sections[sectionName] = sectionContent;
+        const sectionSlug = sectionNameToSlug(sectionName);
+        sections[sectionSlug] = {
+            title: sectionName,
+            content: sectionContent
+        }
```
peterbe: I'm trying to comprehend how we now deal with someone editing `prose.md` and type in `Short Description`. Had we just stuck to `short_description` (or `short-description` depending on the angle) we'd be safe. However, there's definitely intrinsic value in consistency and a breakage here would alert a contributor that we prefer "Short description" instead of "Short Description" and if you don't adhere it will break the rendering.

scripts/build-json/slice-prose.js
```diff
@@ -49,14 +53,18 @@ function packageValues(heading, dom) {
...
     } else {
         const additionalSection = {
-            name: sectionName,
+            title: sectionName,
```
peterbe: Oh, here's where it ends up if someone accidentally typed:
```
## Short Description
```

It could be remedied by matching on the slug instead. Since `'Short Description ==> 'short_description' <== 'Short description'` but it'd only take care of the case and it could potentially "accept" people typing "sHoRt DeScRiPtIoN" too and get away with it.

peterbe:
Not merging yet because I'm interested in hearing some thoughts on my comments from @wbamberg

scripts/build-json/slice-prose.js
wbamberg: Sorry, I'm not sure what point you are making here.

We will validate on the exact text, so "Short Description" !== "Short description" and we will raise an error.

The code doesn't validate input yet, as it doesn't validate any other inputs.

peterbe: I edited the `prose.md` and changed it from `Short description` to `Short Description`. Naughty me. Then ran the script again. No error. Just that instead of being in the main `.prose` key it got put into the `.additional_sections` bucket.

wbamberg: That is what the code here does now, because it doesn't validate yet. As discussed in https://github.com/mdn/stumptown-experiment/issues/19, the proposal is:

As an author, you can:

* use any of the predefined section names ("Short description", "Accessibility concerns", and so on). These are documented somewhere (probably the recipe)

* use a section name of your own invention. If you do this you must register the section name in meta.yaml. This registration is to protect authors from typoes having unexpected results.

We will then:

* if the section name matches a predefined named section, put it there
* if it doesn't, but it matches a registered additional section, put it in additional_sections
* else, raise an error

What this particular PR is trying to do is stabilise somewhat the data structures:

* stabilise the format of prose.md, which enables us to write a validator and to update existing content
* stabilise the format of the JSON, which enables the rendering code to not be broken so often

There is stuff internal to the JSON builder (the validation) that still needs to be done, but hopefully it will not affect those structures.

> Actually, if we do that we could generate the namedSections list instead of hardcoding it.

I'm not sure what this means. Somehow stumptown needs to know what the heading text for the `short_description` section needs to be. How could we generate it, and what would we generate it from?

wbamberg:
> Not merging yet because I'm interested in hearing some thoughts on my comments from @wbamberg

I'd really like to merge https://github.com/peterbe/mdn2/pull/3 before we merge this.

scripts/build-json/slice-prose.js
peterbe: > I'm not sure what this means. Somehow stumptown needs to know what the heading text for the short_description section needs to be

What I mean is that you could read the recipe and extract all the `prose.` sections. E.g. `required: ['short_description', 'usage'], optional: ['accessibility']` 

Now, you render the Markdown and look at the `h2` IDs. I.e. 
```markdown
## Short Description
bla
## Usage and Getting Started
bla
## Accessibility
bla
```
Then extract the `h2` IDs from the HTML and end up with something like this:
```
foundInMarkdown === {
  'short_description': {title: "Short Description", content: ...},
  'usage_and_getting_started': {title: "Usage and Getting Started", content: ...}, 
  'accessibility': {title: "Accessibility", content: ...},
}
```
Now, compare the two:
```javascript
recipe = getRecipe('.')
// recipe == {required: ['short_description', 'usage'], optional: ['accessibility']}

content = []

for (must of recipe.required) {
  if (!foundInMarkdown[must]) {
    throw new Error(`Missing prose section '${must}'`)
  }
  content.push(foundInMarkdown[must])
}
```

What it does is that the recipe's rules are invoked early. It becomes agnostic to the English title and the list of required (and optional) prose section IDs are only written down once in one place. 

Reflection: This could still accidentally allow people to type "Short Description" in one page and "Short description" in another. Perhaps kick that can down the road and make it part of validation (e.g. spelling) or linting.

wbamberg: If I understand you, this would make the first MD file processed the source of truth for section heading text? I think I would prefer for the source of truth to be stumptown.

> the list of required (and optional) prose section IDs are only written down once in one place

I agree very much that this is desirable. Not just IDs but also heading text. The recipe is the obvious place to include both ID and canonical heading text, although I'm reluctant to make it bigger and more complicated.

But anyway, as said, I think this doesn't affect the main purpose of this PR which is to stabilise the prose and built JSON formats .

PR #42, opened 2019-05-30
Switch to yarn

### LINKED ISSUE #15: Switch to yarn
yarn is still much better than npm. It's faster, has support for "resolutions", easier to upgrade packages, ability to check that packages weren't manually added (without using the cli), etc. And from experience feels a lot more predictable and solid than npm. 

Just today, I checked out master and ran `npm install` and it caused [this massive change to `package-lock.json`](https://gist.github.com/peterbe/40b89d13c731b3dd30a5119ea5b14ab2)

peterbe:
Fixes #15

By the way, another better thing about using `yarn` is that when run `yarn run {somenamedscript}` and it fails, it the complete output will only be displayed once whereas with `npm run ...` it would show it twice.


peterbe:
NOTE-TO-SELF; This will break mdn2 because that one assumes to use `npm` in its Makefile.

peterbe:
@wbamberg I'd love to escalate this PR. I believe I've listed enough reasons why `yarn` is better than `npm`. 

By the way, we'd still be using `npm` to publish to `npmjs.com` by the way. That's probably the only reason I still use `npm` in myself.

PR #43, opened 2019-05-30
not depth:false

peterbe:
I'm curious how Travis deals differently with this. Seems that a shallow clone is the default, but perhaps I'm reading the docs wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js: stable
 cache: npm
-git:
-  depth: false
 
 script:
   - npm run test


peterbe:
Docs here: https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth

peterbe:
The default is:
```
git clone --depth=50 https://github.com/mdn/stumptown-experiment.git mdn/stumptown-experiment
```
where as before this PR it was:
```
git clone  https://github.com/mdn/stumptown-experiment.git mdn/stumptown-experiment
```

Meaning, without this PR change we will get ALL history. ...which we don't need, do we?
Leaving it to default (max 50 depth) should be quicker for Travis once it gets much bigger, no?

@Elchi3 am I misinterpreting this?

peterbe:
Another thought is, what if we some day want to generate structure *meta* data. Apart from legacy contributions you could run some fancy git blame history for a directory like https://github.com/mdn/stumptown-experiment/tree/master/content/html/elements/video and sum all the ever contributors across the files within that directory. Smells computationally expensive if you have the *whole* git history you all you need is some disk I/O and patience.

Elchi3:
I think you have this right @peterbe. I'm requesting review from @escattone since it looks he he introduced this setting originally.

peterbe:
Some day, I think it would make sense to calculate the contributors from git history but perhaps that will never be rock solid. Perhaps it would make more sense to maintain a relational database instead somewhere. Then, when building documents we could query that database instead. It would definitely solve the problem of legacy contributors. 

Either way, *today* we don't need the complete git history so I think there's good reasons to go ahead with this PR.

peterbe:
I'm going to self rubberstamp this. If it's not perfect, it's easy to revert.

Issue #38, opened 2019-05-29
Prose end-to-end testing with recipes


peterbe:
We've had at least two errors now where the prose sections are busted. https://github.com/mdn/stumptown-experiment/issues/32 and https://github.com/mdn/stumptown-experiment/issues/37

There's not really anything obviously wrong with the Markdown in those files. Maybe there might be, but that's the point. 

What I would like to have is something that reads the recipes, understands which sections are mandatory and then validates the built JSON files for each of these. 

Let's keep it simple and make an end-to-end test exclusively for this. It'd need to first run the `npm run build-json ...` and then iterate over the produced .json files and do checks including loading in the recipe.

peterbe:
@wbamberg I'm happy to take a stab at this one. But before I do I would appreciate your input on my issue description.

wbamberg:
This will definitely be great to have.

* how stable do the data structures need to be before we do this? I think realistically the structures will continue to change for a while yet as we learn more, and it will be a pain to have to keep updating the validation. At some point we just have to accept that pain, but we should try for a very basic level of stability before we do this. Perhaps after https://github.com/mdn/stumptown-experiment/pull/41 is resolved?

* we need docs for the structures that can be used to write the validation. I have a pretty good idea of what the structures should be, but need to write it down (and in the course of writing it down I'll probably realise I don't have as good an idea as I thought :).

* this becomes essential at the point we are adding substantial amounts of content. We're not at this point yet, but maybe we will be soon.

peterbe:
I'm a huge proponent of never writing unit tests that test the implementation. In this context, it means to NOT write validation code/scripts that nitpick around inside the content. What really matters is; do the scripts (e.g. build-json.js) produce content that matches stuff the recipes require. 

Let's not make something that is over-engineered. I don't know if we need a spec but it would be good to get the ball rolling on *something* that does some validation on the prose. When we discover it's insufficient we can beef it up with more complex checks. 

Last, there are, at the moment, more exciting features to work on that works towards a great MVP for structured content.

peterbe:
By the way, I think this is the best description of an MVP ever: https://twitter.com/jopas/status/515301088660959233?s=20

wbamberg:
[...]

PR #44, opened 2019-05-31
Lint markdown files; adjust headings

Elchi3:
Another large PR. I'm very sorry about this :(

I was happy to see we're going from HTML comments to headings to demarcate sections :tada: 

However, looking into headings and markdown, I was super unhappy with how there are two heading styles (settext and atx). And of course, because linting in this repo is limited still, we have already used both of them. I dislike the settext style because it is very implicit and it doesn't support all heading levels but just h1 and h2. So, we will always have to also use atx style heading as well. 
Now we're also using headings as one of our main feature to demarcate content, so I believe it is time to lint this and be strict about it. So this PR does a bunch of things to help us to meet content expectations.

- Change comments to atx style headings
- Enforce atx heading style per linter
- Add remark-frontmatter, so that the linter understands that we're using frontmatters and that it doesn't confuse it with settext style headers which we disallow.
- remark-lint doesn't cause the build to fail on warnings normally, but I think we should be strict and have it fail, as our project fails and wins with well formatted and structured content (this is the `-f` option).
- remark-lint is quite verbose and always lists passed files, I think we don't need this (this is the `-q` option)
- Add in "remark-preset-lint-consistent". This looks like a sensible package that helps us to avoid further inconsistencies. It brings the consistent heading styles (and this is the reason I added it, but it also has other things to help being consistent)

Somewhat unrelated is fixing all the other linting errors as I went through this, sorry about that, but I didn't want to go through the files twice to fix what the linter currently yells at us.

Oh, one question that I had while doing this: What to do whith empty sections? Do we want to have the headline or not?  We did have empty html comment sections, so I've preserved this and added empty sections that start with a heading. Let me know if we should rather remove them.

### DIFF
--- a/content/css/properties/border-block-start-width/prose.md
+++ b/content/css/properties/border-block-start-width/prose.md
@@ -1,13 +1,13 @@
-<!-- <short-description> -->
+## Short description
+
 The **`border-block-start-width`** [CSS](/en-US/docs/Web/CSS)
 property defines the width of the logical block-start border of an
 element, which maps to a physical border width depending on the
 element's writing mode, directionality, and text orientation.
-<!-- </short-description> -->
 
-<!-- <overview> -->
-It
-corresponds to the
+## Overview
+
+It corresponds to the
 [`border-top-width`](/en-US/docs/Web/CSS/border-top-width),
 [`border-right-width`](/en-US/docs/Web/CSS/border-right-width),
 [`border-bottom-width`](/en-US/docs/Web/CSS/border-bottom-width),
@@ -18,33 +18,31 @@ property depending on the values defined for
 [`direction`](/en-US/docs/Web/CSS/direction),
 and
 [`text-orientation`](/en-US/docs/Web/CSS/text-orientation).
-<!-- </overview> -->
 
-<!-- <example-syntax> -->
+## Example syntax
 ```
 /* <'border-width'> values */
 border-block-start-width: 5px;
 border-block-start-width: thick;
 ```
-<!-- </example-syntax> -->
 
-<!-- <syntax-overview> -->
+## Syntax overview
 Related properties are
 [`border-block-end-width`](/en-US/docs/Web/CSS/border-block-end-width),
 [`border-inline-start-width`](/en-US/docs/Web/CSS/border-inline-start-width),
 and
 [`border-inline-end-width`](/en-US/docs/Web/CSS/border-inline-end-width),
 which define the other border widths of the element.
-<!-- </syntax-overview> -->
 
-<!-- <see-also> -->
--   This property maps to one of the physical border properties:
-    [`border-top-width`](/en-US/docs/Web/CSS/border-top-width),
-    [`border-right-width`](/en-US/docs/Web/CSS/border-right-width),
-    [`border-bottom-width`](/en-US/docs/Web/CSS/border-bottom-width),
-    and
-    [`border-left-width`](/en-US/docs/Web/CSS/border-left-width)
--   [`writing-mode`](/en-US/docs/Web/CSS/writing-mode),
-    [`direction`](/en-US/docs/Web/CSS/direction),
-    [`text-orientation`](/en-US/docs/Web/CSS/text-orientation)
-<!-- </see-also> -->
+
+## See also
+
+- This property maps to one of the physical border properties:
+  [`border-top-width`](/en-US/docs/Web/CSS/border-top-width),
+  [`border-right-width`](/en-US/docs/Web/CSS/border-right-width),
+  [`border-bottom-width`](/en-US/docs/Web/CSS/border-bottom-width),
+  and
+  [`border-left-width`](/en-US/docs/Web/CSS/border-left-width)
+- [`writing-mode`](/en-US/docs/Web/CSS/writing-mode),
+  [`direction`](/en-US/docs/Web/CSS/direction),
+  [`text-orientation`](/en-US/docs/Web/CSS/text-orientation)
--- a/content/css/properties/border-block-start-width/values/border-top-width.md
+++ b/content/css/properties/border-block-start-width/values/border-top-width.md
@@ -1,5 +1,4 @@
-
-# border-top-width
+# `border-top-width`
 
 The width of the border.
 
--- a/content/css/properties/border/prose.md
+++ b/content/css/properties/border/prose.md
@@ -66,5 +66,3 @@ values listed below. The order of the values does not matter.
 
 The border will be invisible if its style is not defined. This
 is because the style defaults to `none`.
-
-## See also
--- a/content/css/properties/border/prose.md
+++ b/content/css/properties/border/prose.md
@@ -1,9 +1,10 @@
-<!-- <short-description> -->
+## Short description
+
 The **`border`** [CSS](/en-US/docs/CSS) property sets an element's
 border.
-<!-- </short-description> -->
 
-<!-- <overview> -->
+## Overview
+
 It's a [shorthand](/en-US/docs/Web/CSS/Shorthand_properties)
 for [`border-width`](/en-US/docs/Web/CSS/border-width),
 [`border-style`](/en-US/docs/Web/CSS/border-style),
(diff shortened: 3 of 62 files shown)


peterbe:
@Elchi3 how does this intersect with https://github.com/mdn/stumptown-experiment/pull/41 ?

peterbe:
Note-to-self; this will conflict with https://github.com/mdn/stumptown-experiment/pull/42 whichever lands first.

Elchi3:
> @Elchi3 how does this intersect with #41 ?

It is supplementary. It applies the change throughout and adds "formatting linting" specifically for headings. Next step is to actually add "content linting" that guarantees certain headings like "Short description" per the recipe rules, etc.

peterbe:
> Next step is to actually add "content linting" that guarantees certain headings

Already filed here: https://github.com/mdn/stumptown-experiment/issues/38

@Elchi3 There, I wrote "I'm happy to take a stab at this one." which you can ignore :)

Nothing's black-and-white but I like this healthy dynamic that is developing. That the MDN content team makes sure stumptown's content is stellar (not just in actual content but also in structured perfection) and the MDN dev team (currently, to be honest, me and Will) then assumes stumptown to be perfect and then runs with that assumption. Seems like a healthy boundary. Mind you, it's not healthy if consumers and producers don't communicate either.

Elchi3:
> Already filed here: #38
> @Elchi3 There, I wrote "I'm happy to take a stab at this one." which you can ignore :)

I didn't see that, but I also didn't start any work on that specifically yet. Feel free to take over.

> Mind you, it's not healthy if consumers and producers don't communicate either.

Sorry about that. Last time I was here, this was still very much an experiment, and for me I just hack away on experiments. Ideally with a somewhat creative flow in which I then don't always ask for permission. Again: sorry.

I didn't participate here in months as I have ambitious OKRs that don't really allow me to lose focus on them. I took some time this week as there was a holiday and I wanted to see where things are before I come to Whistler. As said in my other PRs, feel to ignore my PRs or just take them as inspiration.

peterbe:
> Sorry about that. 

Nothing to be sorry for. I didn't nitpick that the issue was already filed. Just trying to help. 

Also, what I meant about communicating between consumers and producers is that if any human tries to fully understand both sides they'd be stretched too thin. But a little stretching is good. 

I really wish everyone on the whole team gets a chance, before meeting in Whistler, to poke more at stumptown and the renderer since it would genuinely help all discussions if everyone surfaces rather than a select few in a skunkworks bunker.

wbamberg:
Thank you @Elchi3 !

This is a big PR but well-explained and there's nothing I disagree with.

>  I was super unhappy with how there are two heading styles (settext and atx). And of course, because linting in this repo is limited still, we have already used both of them. I dislike the settext style because it is very implicit and it doesn't support all heading levels but just h1 and h2. 
>
> * Change comments to atx style headings
> * Enforce atx heading style per linter

(settext/atx are new to me terms). I agree with all this. I'm going to blame the settext style ones, and the other problems you have fixed here, on Pandoc, again.

I think once we get into migration mode we'll have to have a stern talk with Pandoc.

> @Elchi3 how does this intersect with #41 ?

I guess this PR will break stumptown as it stands, because it needs https://github.com/mdn/stumptown-experiment/pull/41/ to handle the change in section demarcation. And https://github.com/mdn/stumptown-experiment/pull/41/ will probably also conflict with this one, since it also updates the headings in video's prose.md.

And https://github.com/mdn/stumptown-experiment/pull/41/ will *also* break mdn2 because it changes the JSON representation for prose sections.

And also also, https://github.com/peterbe/mdn2/pull/3 still isn't merged, and fixing mdn2 for this breakage will destroy that patch.

So I feel like the order here ought to be:

1. merge https://github.com/peterbe/mdn2/pull/3
2. merge https://github.com/mdn/stumptown-experiment/pull/41/. This then breaks mdn2, so:
3. file a new PR to update mdn2 to understand the new JSON structure
4. once that is merged, merge this PR (having sorted out any conflict from https://github.com/mdn/stumptown-experiment/pull/41/)

Does that sound good?

@Elchi3 , I think this is really great work to update all the section demarcators. In my more optimistic moments I think maybe by the end of the sprint we could have not just one page rendering but 50, and this is the kind of change we need to get there :). It's also great to see you working on the linting.

wbamberg:
> What to do with empty sections? 

We should not allow empty sections. We already allow sections to be optional. If there is a good reason sometimes for a page to omit a section, then we should make the section optional. Allowing empty mandatory sections makes a mockery of the whole thing :).

So in concrete terms, say:

* [`<html>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html) doesn't have  "Overview" - it just goes from short description into attributes.
* [`<ins>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ins) doesn't have "Usage notes"

Do we think these pages are defective? Should we write sections for them? Or are they fine as they are, in which case we should make these sections optional. This is really a content design question, and I'm not sure of the answer (and we should also consult with @chrisdavidmills , who was mostly responsible for designing the [structure of HTML element pages that we are following here](https://docs.google.com/document/d/17R-jyS2WVQ9_OIfErRZWRdPSAVYYauF3wfkJUeG07PI/edit) - although I see now that this does not include "Usage notes", oops).

wbamberg:
>  Next step is to actually add "content linting" that guarantees certain headings like "Short description" per the recipe rules, etc.

Yes. My comment at https://github.com/mdn/stumptown-experiment/issues/38#issuecomment-497506905 captures my view of this. I need to take some time to try to write down what "correctness" is for everything. I wasn't intending to do that in this sprint but perhaps I should after all.

> That the MDN content team makes sure stumptown's content is stellar (not just in actual content but also in structured perfection) and the MDN dev team (currently, to be honest, me and Will) then assumes stumptown to be perfect and then runs with that assumption. Seems like a healthy boundary.

Yes :), except I'm not really in the dev team.

Elchi3:
Thanks for your encouraging words, Will! Makes me want to continue with this when I can find the time.

> I think once we get into migration mode we'll have to have a stern talk with Pandoc.

Yes, this. I haven't looked into pandoc settings a lot, but it looks like we need to dive deeper into it, so that we produce nice and consistent markdown. I think this will also become more clear once we have chosen a flavor, but I also think working more with the markdowns and getting to know the details like settext and atx (which were new terms to me as well, btw) is going to help us a lot making good decisions and deciding on one or another flavor or style.

> We should not allow empty sections.

I think this makes sense and I can remove these in this PR if you like or follow up on it.

> Do we think these pages are defective? Should we write sections for them?

This is indeed a content design question, but however we decide I am so impressed that we could now imagine tooling that would help us enforce updated content design questions. So, say we make them optional for now, but then it turns out that people really have a need for these sections. We could then turn on stricter linting that requires the sections and update the content to meet the criteria. So powerful! The recipes/linter will kind of have our content design explicitly stated!

> And #41 will probably also conflict with this one, since it also updates the headings in video's prose.md.

I think this conflict won't be too bad, happy to resolve that when you land #41 first.

wbamberg:
>  1. merge peterbe/mdn2#3
>  2. merge #41. This then breaks mdn2, so:
>  3. update mdn2 to understand the new JSON structure

These are all done, and I've fixed the merge conflict with this one (and then fixed an error I inadvertently introduced in the process).

So we are able to merge this. But I think we should remove the empty sections first. We won't get errors from them yet of course. Please let me know if you don't have time to do this, and I will do it so this PR can be merged.

Thanks again Florian.

(It sounds like HTML element content design might be a good session for the All-Hands.)

Elchi3:
> These are all done, and I've fixed the merge conflict with this one (and then fixed an error I inadvertently introduced in the process).

Awesome, thanks Will!

> So we are able to merge this. But I think we should remove the empty sections first. We won't get errors from them yet of course. Please let me know if you don't have time to do this, and I will do it so this PR can be merged.

I've just done that. As you say, no validation for this yet, so maybe I missed some, but hopefully not :)

> (It sounds like HTML element content design might be a good session for the All-Hands.)

Yes, seems like a good idea. Would also like further brainstorm on (linting) architecture (see my comment on https://github.com/mdn/stumptown-experiment/issues/38#issuecomment-498604116)

PR #49, opened 2019-06-04
Enforce uniqueness of example titles

### LINKED ISSUE #47: Enforce uniqueness of example titles
See https://github.com/peterbe/mdn2/pull/3#discussion_r289526043. If we want to set IDs for example iframes, and derive them from the slugified title, then we should make sure example titles are unique.

peterbe:
Fixes #47

First all, to test this I messed with the `.md` files temporarily. Looked like this:
Screen Shot 2019-06-04 at 10 54 47 AM

Now you get this:
```
▶ npm run build-json video

> stump@0.1.0 build-json /Users/peterbe/dev/MOZILLA/MDN/stumptown-experiment
> node scripts/build-json/build-json.js "video"

/Users/peterbe/dev/MOZILLA/MDN/stumptown-experiment/scripts/build-json/compose-examples.js:58
        throw new Error(
        ^

Error: Every example's description title must be unique (["generic title"])
    at Object.package (/Users/peterbe/dev/MOZILLA/MDN/stumptown-experiment/scripts/build-json/compose-examples.js:58:15)
    at Object.buildPageJSON (/Users/peterbe/dev/MOZILLA/MDN/stumptown-experiment/scripts/build-json/build-page-json.js:60:33)
    at buildJSON (/Users/peterbe/dev/MOZILLA/MDN/stumptown-experiment/scripts/build-json/build-json.js:37:29)
    at Object.<anonymous> (/Users/peterbe/dev/MOZILLA/MDN/stumptown-experiment/scripts/build-json/build-json.js:42:14)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:785:10)
    at Module.load (internal/modules/cjs/loader.js:641:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:837:10)
    at internal/main/run_main_module.js:17:11
```

There is a more important discussion to be had. We *could* **not** do this and instead put code like this in some other place. Instead of doing it in `compose-example.js` it could be done as an "extension" to `build-json.js`. 

For example, the `buildPageJSON` function (in `build-page-json.js`) has this idea of putting together an Object and then writing that to disk. Within the function it does some checks (arguably not related to the actual content much) and then returns a different error other than 0. For example, in [`buildPageJSON()`](https://github.com/mdn/stumptown-experiment/blob/9b706592a12a1c79f6a6e1ced9e9e7d134f3d0c3/scripts/build-json/build-page-json.js#L60)
it does this:
```javascript
element.examples = examples.package(examplesPaths);
```
We could change that to:
```javascript
element.examples = examples.package(examplesPaths);
validateExamples(element.examples);
```

The same effort needs to be done but it's a matter of code layout where we put that business logic. I think I prefer to do that kinda stuff near and close to the code that generates and raise exceptions as early as possible. After all, if you do throw exceptions you can still rewrite the surrounding tooling to work like this:

```javascript
for (let item of items) {
    try {
        errors += buildPage.buildPageJSON(item);
    } catch(exc) {
        console.warn(`Failed to build ${item}`);
       errors += 1;
    }
}
```

### DIFF
--- a/scripts/build-json/compose-examples.js
+++ b/scripts/build-json/compose-examples.js
@@ -40,7 +40,25 @@ function packageExample(examplePath) {
 }
 
 function package(paths) {
-    return paths.map(packageExample);
+    const rendered = paths.map(packageExample);
+
+    // Do invariance checking after all examples have been packaged.
+    // These are checks that only makes sense to run *after* all
+    // individual examples have been packaged.
+    const normalizedTitles = new Set(
+        rendered.map(example => example.description.title)
+        .filter(title => !!title)
+        .map(title => title.trim().toLowerCase())
+    );
+    // Check that they are all unique
+    if (normalizedTitles.size !== rendered.length) {
+        throw new Error(
+            "Every example's description title must be unique " +
+            `(${JSON.stringify([...normalizedTitles])})`
+        );
+    }
+
+    return rendered
 }
 
 module.exports = {


wbamberg:
I'm on the fence about this. The test looks fine to me.

But I wonder if we should have a unified approach to all the validation we're going to bring to the content, and start with an overall picture of all the things we want to validate (somewhat as @Elchi3 outlined in https://github.com/mdn/stumptown-experiment/issues/38#issuecomment-498604116) before we start implementing it.

peterbe:
> I'm on the fence about this.

You know what, so am I. 
Right here right now I'm slightly in favor of this approach because I think there's value in doing the clever validation *near* the code but I can also see the value in separating them. It'd, for example, make the `compose-example.js` module a "purer". 

If there was a dedicated module just for validation it could be nice since it's relatively easy to jot down all sorts of rules and regulations and it'd be focussed on just that. 

Perhaps punt and sleep on it till a clearer picture comes naturally.

peterbe:
I *think* we resolved this. The whole discussion stemmed from the fact that we use titles to generate IDs and IDs always have to be unique. In https://github.com/mdn/stumptown-renderer/pull/24/files we simply removed the code that puts an ID on every `<iframe>` and we already had a conditional on displaying the example title.

PR #58, opened 2019-06-22
Add first version of a tool to help migrate pages from MDN

wbamberg:
This is the first iteration of a scraper to help migrate MDN docs to stumptown.

Usage:

    npm run scrape-mdn <mdn-url> <destination> scrape-children

`<mdn-url>`
A full URL to an MDN page to scrape.

`<destination>`
Where to put the processed document. This is relative to ./content

`scrape-children`
If 'scrape-children' is given, fetch the immediate children of this page and process them as well.
Store the child pages under a subdirectory of `<destination>`, where the subdirectory name is the parent's name.

For example, if you call:

    npm run scrape-mdn https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding learn/html scrape-children

...this will fetch and process the page at
https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding
and all its immediate children.

The page at 'Multimedia_and_embedding' will
be saved as './content/learn/html/multimedia-and-embedding.md'.

Its children will be saved as pages like ./content/learn/html/Multimedia_and_embedding/images-in-html.md.

The pages are 'processed' on the way. This means that they are:
* converted to JSDOM where various things can be done to them, before going back to HTML
* converted to Markdown
* given some front matter (title and mdn_url)

Of course we could extend the processing to do other interesting things.

### DIFF
--- a/scripts/scraper/scrape-mdn.js
+++ b/scripts/scraper/scrape-mdn.js
@@ -45,13 +45,6 @@ function toMarkdown(html) {
 }
 
 /**
- * Converts a document title into a filename.
- */
-function namify(title) {
-  return title.toLowerCase().replace(/[ ?]/gi, "-");
-}
-
-/**
  * Just writes the doc out where we want.
  */
 function writeDoc(subpath, name, doc) {
@@ -108,7 +101,7 @@ async function processDoc(relativeURL, title, destination) {
   const absoluteURL = baseURL + relativeURL;
   const md = String(await processSingleDocContent(absoluteURL));
   const doc = addFrontMatter(title, relativeURL, md);
-  writeDoc(destination, namify(title), doc);
+  writeDoc(destination, relativeURL.split('/').pop(), doc);
 }
 
 /**
--- a/scripts/scraper/scrape-mdn.js
+++ b/scripts/scraper/scrape-mdn.js
@@ -5,36 +5,18 @@ Usage:
 
 npm run scrape-mdn <mdn-url> <destination> scrape-children
 
-<mdn-url>
-A full URL to an MDN page to scrape.
+<mdn-url>: A full URL to an MDN page to scrape.
 
-<destination>
-Where to put the processed document. This is relative to ./content
+<destination>: Where to put the processed document. This is relative to ./content
 
-scrape-children
-If 'scrape-children' is given, fetch the immediate children of this page and process them as well.
-Store the child pages under a subdirectory of <destination>, where the subdirectory
-name is the parent's name.
-
-For example, if you call:
-
-npm run scrape-mdn https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding learn/html scrape-children
-
-...this will fetch and process the page at
-https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding
-and all its immediate children.
-
-The page at 'Multimedia_and_embedding' will
-be saved as './content/learn/html/multimedia-and-embedding.md'.
-
-Its children will
-be saved as pages like ./content/learn/html/Multimedia_and_embedding/images-in-html.md.
+scrape-children: If 'scrape-children' is given, fetch the immediate children
+of this page and process them as well. Store the child pages under a
+subdirectory of <destination>, where the subdirectory name is the parent's name.
 
 The pages are 'processed' on the way. This means that they are:
 * converted to JSDOM where various things can be done to them, before going back to HTML
 * converted to Markdown
 * given some front matter (title and mdn_url)
-
 */
 
 const unified = require('unified');
--- a/scripts/scraper/scrape-mdn.js
+++ b/scripts/scraper/scrape-mdn.js
@@ -1,0 +1,151 @@
+/*
+A scraper to help migrate MDN docs to stumptown.
+
+Usage:
+
+npm run scrape-mdn <mdn-url> <destination> scrape-children
+
+<mdn-url>
+A full URL to an MDN page to scrape.
+
+<destination>
+Where to put the processed document. This is relative to ./content
+
+scrape-children
+If 'scrape-children' is given, fetch the immediate children of this page and process them as well.
+Store the child pages under a subdirectory of <destination>, where the subdirectory
+name is the parent's name.
+
+For example, if you call:
+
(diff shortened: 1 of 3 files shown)


scripts/scraper/scrape-mdn.js
```diff
@@ -0,0 +1,133 @@
...
+    .use(rehype2remark)
+    .use(stringify)
+    .process(html);
+}
+
+/**
+ * Converts a document title into a filename.
+ */
+function namify(title) {
+  return title.toLowerCase().replace(/[ ?]/gi, "-");
```
peterbe: Consider `.trim().replace(/\s+/g, '-')` so that `foo  bar` becomes `foo-bar` instead of `foo--bar`.

wbamberg:
https://www.npmjs.com/package/remark-stringify has some options we could tweak if we wanted to.

@ddbeck  you might also be interested in this. It's loosely based on your linting approach for short-descriptions, and the idea that if we could lint Wiki docs we could fix errors in the Wiki, so migration could be a more gradual process that improves the Wiki content too. Obviously there's a long way to go with this before that would be possible, but it's a start.

scripts/scraper/scrape-mdn.js
```diff
@@ -0,0 +1,133 @@
...
+function processSingleDocContent(mdnURL) {
+  const mdnRaw = mdnURL + '?raw&macros';
+  return fetch(mdnRaw)
+    .then(result => result.text())
+    .then(text => cleanHTML(text))
+    .then(text => toMarkdown(text));
+}
+
+function getPageJSON(url) {
+  return fetch(url).then(result => result.json());
```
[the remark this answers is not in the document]
peterbe: Isn't the result of this function a JavaScript object from the JSON string parsing?
Actually, I'm not familiar with `JSDOM.fromURL()` but it sounds like you'd give it a URL and have it download then then parse the *HTML*.

peterbe:
I too thought the "cli part" was a bit clunky but I'm also not sure if it matters. This script is never going to be perfect as long as the core peeps can "get to the meat" in some/any way.

peterbe:
Would it be possible to assume that the destination is the dirname of the URL. In lower case. I.e. transform `https://developer.mozilla.org/en-US/docs/Foo/Bar/page_stuff` to `foo/bar`.

wbamberg:
> As an experiment, I wrote a unified plugin to strip BCD tables. It was [very easy](https://gist.github.com/ddbeck/f6ae317aa3d36786286724fffe7a283b). For this particular thing, I could have just as easily done the same with JSDOM but I thought it would be good to exercise the different parts of this, if we're going to be extending it. I was very pleased.
> 

Nice. I think that if we are looking at converting "guide pages" from Wiki->stumptown we will end up doing some weird juggling between `?raw` and `?raw&macros`: the second of these gives you the page after macros are processed, which usually what we want (and eliminates xref macros and stuff like that), but the first gives you the actual macro calls, which you will need to get the original BCD query.

Also if we want to process examples we will need a way to turn a single file input into multiple file output.

> I did notice a couple quirks of the CLI that I'd like to see change in either this PR or a follow-up. First, I'd rather use the real path (e.g., `./content/some/destination` rather than `some/destination`), to get the benefit of tab completion. I tried this from the outset and I got a duplicate `content` directory in my `content` directory.

Done, good suggestion.

> Second, I'd like the `scrape-children` flag to look like a flag (e.g., `--scrape-children` — `args.includes` is your friend here), but this is mostly aesthetic.

Yeah I wanted to do this but could not understand how to pass flags into the script. Apparently you have to do:

    npm run script arg -- -flag

...so I have done that, too.

scripts/scraper/scrape-mdn.js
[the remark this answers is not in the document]
wbamberg: Can you elaborate? I want the response JSON out of this, not a JSDOM object. Are you suggesting I get the JSDOM then serialize it back into JSON?

wbamberg:
> Would it be possible to assume that the destination is the dirname of the URL. In lower case. I.e. transform `https://developer.mozilla.org/en-US/docs/Foo/Bar/page_stuff` to `foo/bar`.

In general: not at the moment, no. For example, MDN has "HTML/Element/link" and we have "html/element**s**/link".

We could as a matter of policy choose to preserve the MDN URL structure in stumptown. But that means we can't also align with the BCD organization, and there are places where the MDN URL structure really doesn't seem ideal (for example, we have everything in CSS directly under "CSS").

So I think I'd prefer to make the destination explicit.

scripts/scraper/scrape-mdn.js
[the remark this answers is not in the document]
ddbeck: Oh, sorry, I misread the function name. Disregard.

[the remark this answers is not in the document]
wbamberg: Ha, I did wonder :).

scripts/scraper/scrape-mdn.js
wbamberg: >  `foo  bar` becomes `foo-bar` instead of `foo--bar`.

As far as I can see `foo bar` becomes `foo-bar` with the existing code?

"Advanced text formatting" -> "advanced-text-formatting"

I do agree that we will probably want to think more about exactly what kind of transformation from title to filename we should have though, I am very likely missing some important cases here.

peterbe: Screen Shot 2019-06-26 at 3 02 07 PM

I only tried in the browser JS. What am I missing?

wbamberg: Oh sorry, I didn't see that your `foo bar` in the original comment had multiple spaces. Sure, I can make this change.

I also wanted to replace "?". for pages like https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML.

But I think in general I didn't think much about this conversion and we will probably hit more difficult cases. Looking at it now it would probably be better to use the last part of the slug "The_head_metadata_in_HTML" rather than the title.

wbamberg: I've updated this to use the slug instead of trying to namify the title, and am going to merge this now as it's got an r+ and this is the only other outstanding issue.

As I said we might want to revisit the question of filenames, but this is quite useful as it is so I'd like to get it in.

PR #61, opened 2019-06-26
end-to-end test in Travis

peterbe:
Context is that after this I'd like to pull in `stumptown-renderer` in CI. There are two interesting things that can come out of that. 

A) We would constantly check the "contract" between content and renderer. In the renderer we are specific about which version of the content we want to use and it's not always `master`. When you're changing that "contract" you might not want to block the PR because the renderer isn't yet ready for this change. So when we do execute the renderer from `stumptown-content`'s CI we probably only want to yield a warning. 

B) If the renderer can render this version of the content we can ship that whole static site to a staging URL on netlify. Then you can fully preview your content changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: node_js
 node_js: stable
-
+cache: npm
 git:
   depth: false
+
+script:
+  - npm run test
+  # Note, in a near future we're going to want to build-json for *everything*
+  # and not just the 'html' section.
+  - npm run build-json html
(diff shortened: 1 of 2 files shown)


peterbe:
Not sure what's going on but Travis did receive the PR and not only that, [it passed](https://travis-ci.org/mdn/stumptown-content/builds/550915649).
Might be related to the renaming of the repo.

peterbe:
Ah! There. Now Travis and GitHub checks seems to work. I didn't do anything. Just pushed more code into the PR.

PR #62, opened 2019-06-26
remove leftovers

peterbe:


peterbe:
Some day we'll set up eslint to do this checking instead.

PR #63, opened 2019-07-01
Document GFM and remark selection

ddbeck:
To close https://github.com/mdn/sprints/issues/1505, I've summarized the decision we made about using GitHub Flavored Markdown and remark. I've written this with the expectation that the document would also document any extensions to GFM we choose to use or build as well. Let me know what you think. Thanks!

### DIFF
--- a/project-docs/markdown.md
+++ b/project-docs/markdown.md
@@ -1,0 +1,47 @@
+# Markdown usage
+
+In the stumptown project, we write content in [GitHub Flavored Markdown](https://github.github.com/gfm/) (GFM).
+
+To transform Markdown into other formats (especially HTML), we use [remark](https://github.com/remarkjs/remark) or its parent library, [unified](https://github.com/unifiedjs/unified). Some legacy code may use other Markdown processors; if you're modifying that code, please replace it with remark and friends.
+
+## Background, or why we chose GFM
+
+We have to specify what variant of Markdown we're using and how we're extending it because Markdown was not originally specifed. To quote the GFM specification:
+
+> John Gruber’s canonical description of Markdown’s syntax does not specify the syntax unambiguously. […] Because there is no unambiguous spec, implementations have diverged considerably. As a result, users are often surprised to find that a document that renders one way on one system (say, a GitHub wiki) renders differently on another (say, converting to docbook using pandoc). To make matters worse, because nothing in Markdown counts as a “syntax error,” the divergence often isn’t discovered right away.
+
+We evaluated several different specifications and implementations of  Markdown. To learn more about this process, read [mdn/sprints#1505](https://github.com/mdn/sprints/issues/1505). What follows is a summary of our conclusions.
+
+We concluded that GFM meets many of our requirements:
+
+* GFM is well specified and there are several implementations of the spec.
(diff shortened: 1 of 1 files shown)


project-docs/markdown.md
```diff
@@ -0,0 +1,47 @@
+# Markdown usage
+
+In the stumptown project, we write content in [GitHub Flavored Markdown](https://github.github.com/gfm/) (GFM).
+
+To transform Markdown into other formats (especially HTML), we use [remark](https://github.com/remarkjs/remark) or its parent library, [unified](https://github.com/unifiedjs/unified). Some legacy code may use other Markdown processors; if you're modifying that code, please replace it with remark and friends.
```
peterbe: What "legacy code"?

peterbe: Also, *why* switch between one and another? Why aren't all `.md` files transformed with just one library?

ddbeck: There are several places [where we're not using unified/remark](https://github.com/mdn/stumptown-content/search?q=require%28%27marked%27%29&unscoped_q=require%28%27marked%27%29).

As for the question of two libraries: remark is the Markdown subset of the unified "library" (i.e., the name of the project which includes all the tiny npm packages that make up unified). unified has various tools for handling Markdown-ish content, including the Markdown itself (remark), an abstract representation of the Markdown (mdast), prose (retext), and the resulting HTML (rehype)—it's not so much switching between two libraries as it is using one of the many, many little packages in the unified project to do whatever it is you need to do to Markdown or the resulting HTML.

wbamberg: > There are several places where we're not using unified/remark.

I think @Elchi3 was looking into eliminating this although I don't know how far he got (I'd be happy to take this over if you would like, Florian, since it was my fault in the first place!). But I think the way this in worded:

> Some legacy code may use other Markdown processors; if you're modifying that code, please replace it with remark and friends.

...will age weirdly: there is only a little legacy code, and we should I think just fix that rather than document its existence. So it might be better to express this more in terms of preventing future changes from introducing different parsers, like:

> When reviewing PRs to this repository, check that they do not introduce dependencies on other Markdown parsers.

?

ddbeck: @wbamberg it's a good point that the legacy text won't be meaningful before long. I've added an imperative against introducing other parsers. 👍

ddbeck:
@wbamberg yes, I definitely agree that this doc could (and should) contain our Markdown style decisions. I thought that while writing my initial draft, but couldn't think of a style choice we had already committed to. The headings style is a good one! That said, it's not really critical to mdn/sprints#1505, so I'll open a new PR after this one is merged.

project-docs/markdown.md
Elchi3: > I think @Elchi3 was looking into eliminating this although I don't know how far he got (I'd be happy to take this over if you would like, Florian, since it was my fault in the first place!). 

I could probably resurrect this work, but feel free to beat me to it. It was mostly bugging me that we use `marked` and `remark` at the same time. We should just stick to one library here and I think (and as Daniel has found out) the `unified` library (or rather its universe of packages) looks well maintained and suitable for stumptown purposes.

PR #70, opened 2019-07-11
Add 'reference' subdirectory

wbamberg:
This PR moves:

* content/css/properties -> content/css/reference/properties
* content/html/elements -> content/html/reference/elements

...as discussed in https://github.com/mdn/stumptown-content/wiki/Notes-from-Whistler-All-Hands%2C-June-2019#contenthtmlelements---contenthtmlreferenceelements.

I think this might break the renderer because it also changes the output directory structure to match, and the renderer might hardcode paths. So we will need a follow-up there.


scripts/build-json/build-page-json.js
```diff
@@ -17,7 +17,7 @@ const writeToFile = (propertyName, json) => {
     }
   };
 
-  const dest = path.join(process.cwd(),'packaged/html/elements', `${propertyName}.json`);
+  const dest = path.join(process.cwd(),'packaged/html/reference/elements', `${propertyName}.json`);
```
peterbe: I should have made a point about this before but the use of `/`, will that work for contributors on Windows?

peterbe:
stumptown-renderer is fragile because it hardcoded the [object lookup](https://github.com/mdn/stumptown-renderer/blob/1a2b0eaba0dcd22c13b714671e7edaf9ff94eca1/cli/index.js#L50)

I can easily fix that by splitting the `filePath` variable and loop over it. But I'd **rather** make the switch to:

```javascript
  const options = {
    document: jsonData.data,
    metadata: jsonData.metadata,
  };
```
I.e. change `video.json` to look like this:
```
▶ head video.json
{
  "meteadata":  {
      "contributors": [...],
      "modified": ...,
  },
  "data": {
      "title": "<video>: The Video Embed element",
  ...
```

Hint hint

wbamberg:
> I can easily fix that by splitting the filePath variable and loop over it. But I'd rather make the switch to:

I wrote a PR for this one yesterday but wasn't sure what to put in `metadata`. But I guess contributors is a good one. We don't have modified date supported yet. Can you think of anything else? If we start using recipes to drive the renderer, is that data or metadata?

scripts/build-json/build-page-json.js
wbamberg: I just pushed a commit which should remove this bit of hardcoding and just ensure we use the same directory structure under /packaged as we found under /content. Although I'm not particularly in love with all the path-fiddling going on in there, and there's probably a much nicer way to deal with that.

peterbe:
> We don't have modified date supported yet. 

That's a shame. I've always liked looking at that in MDN. 

>  But I guess contributors is a good one

Yes!
Do you mind tackling turning it into non-HTML. E.g. a list of dicts. 

> Can you think of anything else?

Yes, the recipe. 

Can't think of anything else (that we can't add later) at the moment.

scripts/build-json/build-page-json.js
peterbe: Yes, it's a fiddly at the moment. 
Perhaps because we can't see it just yet but there must be some nifty function that we can use/write that just takes care of all of this in one sweep. Basically, something that changes the root and takes care of creating the sub-directories etc.

Issue #68, opened 2019-07-11
Should recipes drive the renderer?


wbamberg:
(meta: taken from https://github.com/mdn/stumptown-content/wiki/Notes-from-Whistler-All-Hands,-June-2019#decide-whether-recipes-should-drive-the-renderer)

In stumptown we've expressed a desire to distinguish between (1) "MDN content" which is an abstract collection of web documentation, and (2) "MDN the website" (developer.mozilla.org). The idea is that "MDN the website" is the main consumer of "MDN content", but not necessarily the only one: editors and devtools could also want to integrate the content. This is reflected in the division between "stumptown-content" and "stumptown-renderer".

Maintaining this distinction is tricky, and one of the skirmishes is the meaning of "recipes". These describe what content should be included in a particular type of document (e.g. https://github.com/mdn/stumptown-content/blob/master/recipes/html-element.yaml). These files have - or at least used to have - a sort of double life. On the one hand they describe which bits of content authors should supply when they're documenting a thing. On the other hand they might be a description of how MDN should render pages.

The concrete difference would be seen if, for example, we decided to stop displaying CSS formal syntax in pages, but wanted to keep the content (maybe because other consumers were using it).

We discussed this recently (https://github.com/mdn/stumptown-content/issues/48) and the resolution was that recipes should *not* drive the renderer, but instead we should just hardcode the different content types are React components (https://github.com/mdn/stumptown-renderer/pull/18, which basically implements https://github.com/mdn/stumptown-content/issues/48#issuecomment-498992233). This means that the renderer has nothing to do with recipes.

In Whistler, @peterbe wanted us to reconsider this decision, and make recipes drive the renderer.

Pros: things get simpler. Not just in the case of recipes: there are other places too where this distinction can get messy: an obvious one is sidebars, which seem a very MDN-y thing, but whose content is clearly an authorial choice (so it is very tempting to put them in stumptown-content). It's easier to make design choices to satisfy real consumers than to satisfy theoretical consumers.

Cons: we acknowledge that stumptown-content is being built to serve MDN-the-website. It's still possible for other consumers to use it, and it might still be as easy, but it's certain that we will introduce extra stuff they don't care about (like sidebars) and possible we will make choices about what to expose or how to expose it that are very suboptimal for other consumers.

Either way, we should be very clear about the choice we're making here and its implications.

peterbe:
Coupla thoughts...

[...]

PR #73, opened 2019-07-15
Remove wrapping from built JSON

wbamberg:
This PR remove the "html.elements.*" wrapping from the built JSON.

It also splits the content between `data` and `metadata`. As discussed in https://github.com/mdn/stumptown-content/pull/70#issuecomment-510909281 and following, `metadata` currently contains just `contributors` but might also include `recipe` depending on how people feel about https://github.com/mdn/stumptown-content/issues/68, and might also include `modified_data` if we want to support that.

### DIFF
--- a/scripts/build-json/build-page-json.js
+++ b/scripts/build-json/build-page-json.js
@@ -39,28 +39,25 @@ async function buildPageJSON(elementRelativePath) {
     const prosePath = path.join(elementPath, 'prose.md');
     const contributorsPath = path.join(elementPath, 'contributors.md');
 
-    // set up element data
-    const element = {
-      data: {},
-      metadata: {}
-    };
-    element.data.title = meta.title;
-    element.data.mdn_url = meta['mdn-url'];
-    element.data.interactive_example_url = meta['interactive-example'];
-    element.data.browser_compatibility = bcd.package(meta['browser-compatibility']);
+    // set up data
+    const data = {};
+    data.title = meta.title;
+    data.mdn_url = meta['mdn-url'];
+    data.interactive_example_url = meta['interactive-example'];
+    data.browser_compatibility = bcd.package(meta['browser-compatibility']);
     if (meta.attributes['element-specific']) {
         const attributesPath = path.join(elementPath, meta.attributes['element-specific']);
-        element.data.attributes = await attributes.package(attributesPath);
+        data.attributes = await attributes.package(attributesPath);
     } else {
-        element.data.attributes = [];
+        data.attributes = [];
     }
-    element.data.examples = await examples.package(examplesPaths);
-    element.data.prose = await prose.package(prosePath);
+    data.examples = await examples.package(examplesPaths);
+    data.prose = await prose.package(prosePath);
 
     // set up element metadata
-    element.metadata.contributors = await contributors.package(contributorsPath);
+    data.contributors = await contributors.package(contributorsPath);
 
-    writeToFile(element, elementRelativePath);
+    writeToFile(data, elementRelativePath);
     console.log(`Processed: ${elementRelativePath}`);
     return 0;
 }
--- a/scripts/build-json/build-page-json.js
+++ b/scripts/build-json/build-page-json.js
@@ -8,23 +8,15 @@ const attributes = require('./compose-attributes');
 const prose = require('./slice-prose');
 const contributors = require('./resolve-contributors');
 
-const writeToFile = (json, elementPath) => {
+function writeToFile(json, elementPath) {
   const propertyName = path.basename(elementPath);
   const dirName = path.dirname(elementPath);
-  const data = {
-    html: {
-      elements: {
-        [propertyName]: json,
-      }
-    }
-  };
-
   const dest = path.join(process.cwd(), 'packaged', dirName, `${propertyName}.json`);
   const destDir = path.dirname(dest);
   if (!fs.existsSync(destDir)) {
       fs.mkdirSync(destDir, { recursive: true });
   }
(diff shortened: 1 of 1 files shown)


scripts/build-json/build-page-json.js
```diff
@@ -8,23 +8,15 @@ const attributes = require('./compose-attributes');
 const prose = require('./slice-prose');
 const contributors = require('./resolve-contributors');
 
-const writeToFile = (json, elementPath) => {
+function writeToFile(json, elementPath) {
```
peterbe: Thank you.

scripts/build-json/build-page-json.js
```diff
@@ -47,21 +39,26 @@ async function buildPageJSON(elementRelativePath) {
...
-    element.mdn_url = meta['mdn-url'];
-    element.interactive_example_url = meta['interactive-example'];
-    element.browser_compatibility = bcd.package(meta['browser-compatibility']);
+    // set up element data
+    const element = {
+      data: {},
+      metadata: {}
+    };
+    element.data.title = meta.title;
+    element.data.mdn_url = meta['mdn-url'];
```
peterbe: Awesome! I can take care and followup on this in the renderer.

scripts/build-json/build-page-json.js
```diff
@@ -47,21 +39,26 @@ async function buildPageJSON(elementRelativePath) {
...
-    element.title = meta.title;
-    element.mdn_url = meta['mdn-url'];
-    element.interactive_example_url = meta['interactive-example'];
-    element.browser_compatibility = bcd.package(meta['browser-compatibility']);
+    // set up element data
+    const element = {
+      data: {},
+      metadata: {}
+    };
+    element.data.title = meta.title;
```
peterbe: Why is GitHub's javascript syntax highligher lighting up on the `data` key thing but not on `metadata`? It's not a builtin as far as I know.

scripts/build-json/build-page-json.js
```diff
@@ -47,21 +39,26 @@ async function buildPageJSON(elementRelativePath) {
...
     const contributorsPath = path.join(elementPath, 'contributors.md');
 
-    // make the package
-    const element = {};
-    element.title = meta.title;
-    element.mdn_url = meta['mdn-url'];
-    element.interactive_example_url = meta['interactive-example'];
-    element.browser_compatibility = bcd.package(meta['browser-compatibility']);
+    // set up element data
+    const element = {
```
peterbe: Does this variable name make sense any more?

scripts/build-json/build-page-json.js
ddbeck: But why? I'm not clear on what this has to do with either of the issues here

peterbe:
Yeah actually, the renderer doesn't actually benefit from a distinction of `data` and `metadata`. The renderer needs `mdn_url`, `contributors`, `last_modified` (doesn't exist yet), `recipe`.

scripts/build-json/build-page-json.js
peterbe: Probably nothing but I'd be willing to accept this "violation". The project is young but once it matures we'll have to become stricter.

wbamberg: You're right, it doesn't. I will put it in a separate PR if you prefer.

scripts/build-json/build-page-json.js
wbamberg: No, but this whole thing will need redoing at some point to be not just about HTML elements, and I don't think this is the point for that.

wbamberg:
`data`/ `metadata` was Peter's suggestion, but I also wasn't sure it was an easy distinction to make. I'm happy to move `mdn_url` there, or to abandon the distinction and put everything in the top level.

peterbe:
> `data`/ `metadata` was Peter's suggestion

Yeah, and I was slow to realize that there's very little data that is *not* needed by the renderer. 

Another thing I've been slow to realize is how unimportant this is. If we one day decide to rearrange this, we can just make that change immediately (in stumptown-content) and in stumptown-renderer we just make one patch that changes the git submodule pointer and some logic in the `cli/index.js` and voila it uses the new structure. 

So, let's just go ahead and keep it dead simple. 
```javascript
const data = {};
data.mdn_url = meta['mdn-url'];
data.examples = await examples.package(examplesPaths);
data.contributors = await contributors.package(contributorsPath);
data.interactive_example_url = meta['interactive-example'];
data.prose = await prose.package(prosePath);
...
...

fs.writeFileSync(dest, `${JSON.stringify(data, null, 2)}`);
```

wbamberg:
I have removed the `data`/`metadata` level. As you say it's easy enough to add later if we decide it will be useful.

scripts/build-json/build-page-json.js
ddbeck: Is this a style thing? I'm cool with grab bag PRs—everything's in flux right now—I just literally didn't know why you changed it from an arrow (I didn't spot `this` or anything, though I suppose I missed something). I was just curious :-)

wbamberg: > Is this a style thing? 

Yes (sorry to be confusing). I'm much more familiar with using named functions than assigning anonymous functions to variables. So that's what this script does, mostly, but I copied this bit from your short-descriptions linter, and that used the other style. I though it was weird to have a mixture, and since I was editing this function anyway, thought I'd update it to be in line with the rest.

ddbeck: Cool, thank you for the explanation, Will!

ddbeck:
Three green check marks. I'll merge this now. 🎉

Issue #55, opened 2019-06-12
Clarify required and optional prose sections for HTML elements


wbamberg:
The recipe for HTML elements lists the following as mandatory sections:

- Short description
- Usage notes
- See also

and the following as optional sections:

- Overview
- Attributes text (this is for elements that want to add some extra prose for the attributes list)
- Accessibility concerns

...with of course the option of providing any extra sections that are specific to a particular element. This doesn't quite match up with the existing stumptown content (as scraped from MDN).

It also doesn't quite match up with the specification @chrisdavidmills  wrote for HTML docs: https://docs.google.com/document/d/17R-jyS2WVQ9_OIfErRZWRdPSAVYYauF3wfkJUeG07PI/edit#heading=h.nlsh2gn0fxac:

> Summary/intro text

OK, call that "Short description".

> Additional usage information; a full explanation of how to use the element. 

Call that "Overview".

> Styling with CSS

We don't have this at all.

> Accessibility concerns and best practices

We have this as "Accessibility concerns".

> See also

We have this of course.

So: the doc adds "Styling with CSS", which seems like a useful section at least sometimes, and also omits "Usage notes", which seems like a good omission (we would be better to fold that content into "Overview"). So suggest we could change the recipe to:

**Mandatory sections:**

- Short description
- See also

**Optional sections:**

- Overview
- Styling with CSS
- Attributes text
- Accessibility concerns

I don't know if we should be more prescriptive than this, and perhaps make "Overview" or even "Styling with CSS" mandatory.

For "Overview": there are existing elements which don't have an overview, so if we make it mandatory we have to update them. For some elements we might want to move the Usage notes to Overview (e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i#Usage_notes).

For "Styling with CSS" there will of course be some elements where it's not applicable at all, and others where there's nothing very interesting to say. So this probably should be kept optional.

@chrisdavidmills , thoughts?

---

# Acceptance criteria
- [x] Recipe for HTML elements is complete
- [x] Existing content that would not comply with the recipe has been identified
- [x] Issues have been opened for updating content, updating build-json, adding linting, and updating the renderer

# Tasks 
(see https://github.com/mdn/stumptown-content/issues/55#issuecomment-510422655 for details):

- [x] update the recipe
- [x] identify existing content that would not comply with the recipe
- [x] open issue(s) for updating content, updating build-json, adding linting, and updating the renderer

chrisdavidmills:
[...]

PR #74, opened 2019-07-23
Update required and optional sections in HTML recipe

ddbeck:
This PR implements the recipe changes discussed in #55.

One area of uncertainty here is the order of the styling with CSS section. This is a newly defined section and I couldn't find an example of one of these in the wild. My ordering is basically a guess.

### DIFF
--- a/recipes/html-element.yaml
+++ b/recipes/html-element.yaml
@@ -5,7 +5,6 @@ body:
 - prose.overview
 - prose.attributes_text?
 - meta.attributes
-- prose.usage_notes
 - prose.*
 - prose.styling?
 - prose.accessibility_concerns?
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks @ddbeck ! I think we also decided to remove `usage_notes` and fold it into `overview` (https://github.com/mdn/stumptown-content/issues/55#issuecomment-501626977).

On order, perhaps it would be best to have all the named sections (except `see_also`), then all the custom sections?

Like:

```
related_content: /content/related_content/html.yaml
body:
- prose.short_description
- meta.interactive_example?
- prose.overview
- prose.attributes_text?
- meta.attributes
- prose.styling?
- prose.accessibility_concerns?
- prose.*
- meta.examples
- meta.info_box:
    - meta.api
    - meta.permitted_aria_roles
    - meta.tag_omission
- meta.browser_compatibility
- prose.see_also
```

...it just seems like a simpler rule that way.

ddbeck:
Thanks, @wbamberg! Done.

Issue #69, opened 2019-07-11
Should we use front matter instead of meta.yaml?


wbamberg:
(meta: taken from https://github.com/mdn/stumptown-content/wiki/Notes-from-Whistler-All-Hands,-June-2019#front-matter-versus-metayaml-files)

In the HTML element docs we have prose.md for the content and meta.yaml for metadata, which includes `title` and `mdn_url` but also things like `browser_compatibility`.

In the guide pages we also have some metadata: at least `title` and `mdn_url`. We seem to lean here towards keeping them in the Markdown file as front matter:

```
---
title: Introduction to HTML
mdn_url: https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML
---
At its heart, HTML is a fairly simple language...
```

This seems to be inconsistent. Should we use front matter for `title` and `mdn_url` even in reference docs? Should we keep all metadata in front matter, and abolish meta.yaml entirely?

We have had this conversation before, and decided to keep it in a separate file, for now. The argument for that being that lots of metadata at the start of a Markdown file makes it harder to read. But we should rethink this.

wbamberg:
Especially given the stuff about guide pages above, I think I'm in favour of using front matter here. Having fewer files and keeping more stuff together is also a point in favour of front matter.

peterbe:
I think the best argument I can think of is that it looks just fine in GitHub's preview thingy. 

<img width="1040" alt="Screen Shot 2019-07-11 at 4 03 54 PM" src="https://user-images.githubusercontent.com/26739/61081585-8c705d80-a3f5-11e9-93a7-57844b830277.png">

Also, if you're in a file called something super generic like `prose.md`, in your editor, the having the front-matter there would actually be super helpful. E.g.:

<img width="767" alt="Screen Shot 2019-07-11 at 4 05 07 PM" src="https://user-images.githubusercontent.com/26739/61081667-c477a080-a3f5-11e9-9e45-9b186a9e88e5.png">

peterbe:
Less files ==> 👍

Issue #90, opened 2019-07-30
Write linter to validate that pages follow their recipes


ddbeck:
As a follow up to #55 and #74, we ought to have a script that will error out when when pages have known deprecated sections (like HTML's `usage_notes`) and when pages are missing required sections (like `overview`).

Acceptance criteria:
- [ ] Running `npm run lint` runs this script
- [ ] Errors result if any deprecated sections appear in a page
- [ ] Errors result if any page is missing a required section

wbamberg:
@ddbeck , I wasn't sure if you were planning to work on this or not, but either way: for https://github.com/mdn/stumptown-content/pull/91 I found myself making lots of changes to the content, and wrote a script to help me be confident I wasn't missing anything.

So I thought it might be useful to contribute it somehow to more general content linting efforts, so I pushed it to this branch: https://github.com/wbamberg/stumptown-experiment/blob/test-content/scripts/test/test-content.js (note that this is on top of https://github.com/mdn/stumptown-content/pull/91). The way it's invoked would want to change, and it could surely use some refinement in terms of how it reports errors, but I thought at least it's worth a look.

The idea is that it uses recipes to test content: so it looks for content files, looks up their recipe, then checks that all the things required in the recipe are present in the content. It could eventually go much deeper than this, and check, for example, that live samples listed are all properly formatted.

ddbeck:
Thank you, @wbamberg. That's really helpful. I also wrote a similar script for https://github.com/mdn/stumptown-content/issues/55#issuecomment-515417887. We used a similar pattern to each other of walking the content, loading the recipe, and checking for required sections (mine also checks for deprecated sections). Once #91 is merged, I would like to come back to this.

ddbeck:
I opened https://github.com/mdn/stumptown-content/pull/169 to implement the acceptance criteria for this issue.

PR #91, opened 2019-07-31
Meta to front matter

wbamberg:
Fixes https://github.com/mdn/stumptown-content/issues/69.

Sorry this is such a mega-PR. I was going to change just video's content and the script, and then change everything else later. But then I realized `related_content` would not work properly without updating everything.

So this change should:
* delete `meta.yaml` for all files under html/reference/elements
* move the contents of `meta.yaml` into `prose.md` as front matter, and renamed the result to `docs.md`
* updated all the build-json code to reflect this change

I have also updated all keys in front matter to use `_` rather than `-`: so, `mdn-url`->`mdn_url` etc. This wasn't needed but while I was in there it seemed like a good point to make this change.

I have verified that this produces identical JSON output to the old code.

I have also written a little test tool that does some basic validation of docs.md against its recipe. I used this to catch errors in this PR, but didn't include the tool itself in the PR. Perhaps it could be a part of https://github.com/mdn/stumptown-content/issues/90?

This work also made me think: should we include `title` and `mdn_url` in the html-element recipe?


content/html/reference/elements/abbr/docs.md
```diff
@@ -1,3 +1,24 @@
+---
+title: '<abbr>: The Abbreviation element'
+mdn_url: https://developer.mozilla.org/docs/Web/HTML/Element/abbr
+tags:
+    group: Inline text semantics
+api: HTMLElement
+permitted_aria_roles:
+    - any
+tag_omission: none
```
peterbe: What are all these (`tags`, `api`, `permitted_aria_roles`, `tag_omission`)??

By the way, in the new read-only MDN site (e.g. beta.developer.mozilla.org) we have removed all display of tags.

peterbe:
> This work also made me think: should we include `title` and `mdn_url` in the html-element recipe?

Nah. I think you can hardcode those in the validator/linter.

content/html/reference/elements/abbr/docs.md
wbamberg: `api`, `permitted_aria_roles`, `tag_omission` are all things that live in the "technical summary" box (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#Technical_summary).

We didn't think too hard about these when we added them to the recipe, and I'm not sure if we really want all of them to be represented as data (yes on `api` I think, and perhaps `permitted_aria_roles`, but `tag_omission` is tricky to represent structured). Also I think we might want a better name than `api` - it's called "DOM Interface" in the page.

`tags` is just generic tags. We don't have a use for tags yet but I expect we might in future.

> By the way, in the new read-only MDN site (e.g. beta.developer.mozilla.org) we have removed all display of tags.

\o/. Tags in Kuma are a [horrible mess](https://developer.mozilla.org/en-US/docs/tags).

peterbe: I suppose we, like with examples, make that a `tech_summary.yaml` file in the `html/reference/element/video/` directory and have a React "ingredient" component to render it out. But not a blocker for progress.

peterbe:
>  I'd love to see a move to `tagname.md` instead of `prose.md` or `docs.md`

I agree. I hate when you have many folders, in a javascript project, all filled with `index.js` files and your IDE tabs become less useful. But if it's a hassle don't let it block.

wbamberg:
> This seems fine to me, though I'd love to see a move to `tagname.md` instead of `prose.md` or `docs.md`

This is not as simple as it seems. Currently the build-json and linting code code [rely on specific filenames to find the things they need to build/lint](https://github.com/mdn/stumptown-content/blob/master/scripts/build-json/build-json.js#L10). We can't just process every .md file under /content because there are also "supplementary files" like attribute descriptions and example descriptions.

So I guess (mostly unpalatable) options include:

* continue to use magic names for files. This has the drawback you pointed out and it also fragile and implicit: if someone happens to call a different file docs.md then things will break.

* only allow these special kinds of .md files under /content, and move other .md files somewhere else (so for instance the attribute descriptions could live somewhere outside /content, and the files in /content could point to them). I think this would be unfortunate for authors - it's a nice feature of the current organization that all the supplementary files are kept together with the main file.

* don't have supplementary files. Keep everything for a given element in one file, including live samples (not interactive examples of course). This would make these files much bigger and would give us more complicated rules for slicing them up, and would make it harder to reuse content in some ways (since 2 HTML elements, say, couldn't share an example). But it would make other things simpler, and we'd have much fewer files that way of course.

* don't scan for what to build by just walking the tree looking for things. Instead, use something like the "related_content" to tell you what to build. So the build-json starts by thinking "I want to build the HTML docs", so it finds https://github.com/mdn/stumptown-content/blob/master/content/related_content/html.yaml, opens it, resolves it, and walks through the tree building everything in there. I think I like this because it's very explicit, not relying on any kind of magic, and it also guarantees that the tree of built pages, the sidebar, and any index pages are always in sync. Note also that even in the very limited html.yaml file already in there, there are docs (the learning area docs) that aren't under the content/html bit of the tree, but that are needed if you want a complete set of MDN's HTML docs.

ddbeck:
Yeah, there's enough complexity in the file naming thing to not do it in this PR (I kinda thought `<dirname>/<dirname>.md` wouldn't be that big a deal, but the more I think on it, the more issues there are). Let's move on without that for now.

wbamberg:
> Yeah, there's enough complexity in the file naming thing to not do it in this PR (I kinda thought `<dirname>/<dirname>.md` wouldn't be that big a deal, but the more I think on it, the more issues there are). Let's move on without that for now.

OK, sounds good. But I think we will have to face this monster at some point!

ddbeck:
Opened #93 to address the file name question.

Issue #46, opened 2019-05-31
Represent the height of an interactive example


wbamberg:
With interactive examples you can have different heights of example, to support examples that need more code (like [`<colgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/colgroup)) or those that need less (like [`<i>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i).

We need a way for the author to represent this.

******

**Acceptance criteria:**
* stumptown-content machinery can understand interactive example height indicator
* content itself includes the appropriate value for height
* the renderer interprets the height and creates an iframe of the appropriate size.

wbamberg:
Fixed by https://github.com/mdn/stumptown-content/pull/135 and https://github.com/mdn/stumptown-renderer/pull/129.

PR #98, opened 2019-08-07
Fix #92 - related.buildRelatedContent should be memoized

### LINKED ISSUE #92: related.buildRelatedContent should be memoized
If you, for example, run `npm run build-json html` it will call `buildRelatedContent` with the argument `'/content/related_content/html.yaml'` for every html ref page.
```
item.related_content = related.buildRelatedContent(recipe.related_content);
```

joedarc:
Before this patch, `npm run build-json html` called the `buildRelatedContent` function 50 times with the same argument. This does `readFileSync` each time, which is expensive. This patch caches the result of `buildRelatedContent` depending on the argument, and returns the cache if it's been set.

### DIFF
--- a/scripts/build-json/related-content.js
+++ b/scripts/build-json/related-content.js
@@ -72,7 +72,7 @@ function buildSection(sectionSpec) {
 const relatedContentCache = {};
 function buildRelatedContent(specName) {
   const cached = relatedContentCache[specName];
-  if (cached) {
+  if (cached !== undefined) {
     return cached;
   }
   const spec = yaml.safeLoad(fs.readFileSync(path.join(process.cwd(), specName), 'utf8'));
--- a/scripts/build-json/related-content.js
+++ b/scripts/build-json/related-content.js
@@ -69,9 +69,16 @@ function buildSection(sectionSpec) {
  * Build a single related content object given its YAML file.
  * At the top level a related content object is an array of sections.
  */
+const relatedContentCache = {};
 function buildRelatedContent(specName) {
+  const cached = relatedContentCache[specName];
+  if (cached) {
+    return cached;
+  }
   const spec = yaml.safeLoad(fs.readFileSync(path.join(process.cwd(), specName), 'utf8'));
-  return spec.map(buildSection);
+  const result = spec.map(buildSection);
+  relatedContentCache[specName] = result;
+  return result;
 }
 
 module.exports = {


scripts/build-json/related-content.js
```diff
@@ -69,9 +69,16 @@ function buildSection(sectionSpec) {
  * Build a single related content object given its YAML file.
  * At the top level a related content object is an array of sections.
  */
+const relatedContentCache = {};
 function buildRelatedContent(specName) {
+  const cached = relatedContentCache[specName];
+  if (cached) {
```
peterbe: shouldn't that be `if (cached !== undefined)` or something? In case, the outcome of `spec.map(buildSection);` is something falsy.

joedarc: Makes sense @peterbe, updated to check for undefined. 👍

joedarc:
Thank you for the feedback @wbamberg! I definitely understand this project is an experiment, just figured I'd get #92 out of the way since it was a quick fix. Also, makes a lot of sense that the related content should pertain to what you are viewing.

Having said that, are there any more pressing issues you'd like me to work on? I'm looking to contribute and wanted to know if there's a list of issues anywhere else like in Bugzilla that have a higher priority. And would you like to close or merge this PR?

wbamberg:
Let's merge it :).

>  are there any more pressing issues you'd like me to work on? 

For this project:

* I think it would be great to have better testing/linting, as this helps us when we want to make changes to the document structures (which we end up doing quite a lot of). There's https://github.com/mdn/stumptown-content/issues/90 for this, although I don't know if @ddbeck wants to take it himself or if he would be happy to review/mentor patches.

* https://github.com/mdn/stumptown-content/issues/46 is quite well-defined I think and will help the rendered output look better.

* over in [stumptown-renderer](https://github.com/mdn/stumptown-renderer/), there's https://github.com/mdn/stumptown-renderer/issues/65, and also we're not rendering BCD tables yet. I just filed https://github.com/mdn/stumptown-renderer/issues/69 for this. I also have a half-completed (well not even half-completed, really :) branch for this, that I could push somewhere in case it helps.

peterbe:
>     * does it matter that it's 100x slower to build sidebars?

True, and a counter-argument against merging or even working on this. However, what I care about is "Does it scale?" Not all of them but some of those kinds of questions definitely belong in the experiment phase.

For example, for the JSON files in `stumptown-content/packaged/**` the `related_content` is more than half the weight when it's always the same for every page in a certain pattern. For 10,000 pages that would really mean something and squeezing a few milliseconds would be miniscule compared to a broader and "smarter" solution where a host of pages can reuse the same sidebar or something like that. 

I guess I'm trying to defined premature optimization by calling it pre-premature optimization :)

I think we should land lots of fun code but whenever we write something rough and we can drop in a `// TODO: this is ugly and slow but it'll work for now`. So perhaps the order of importance is a: 1) the feature 2) a code-comment that it can/needs to be optimized 3) optimized.

ddbeck:
@wbamberg @joedarc for #90, I am interested in doing the linting stuff but maybe it's a bit of a jerk move to call dibs on something that's not actually in my sprint tasks. I'd be happy to review patches on it, but maybe it would help if I wrote out in more detail first what linting might look like.

@peterbe I think it's fine to be concerned about performance, as long as it's attached to genuine user stories. I think it'd be great to set out some performance expectations for user and developer experience (e.g., as an author, I want to preview a single page in less than Xms; as a reviewer, I want linting and other automated checks to finish in Y minutes; as an MDN dev, I want to be able to regenerate all the content from a cold start in less than Z minutes, etc.). But I don't think we should pursue optimizations just because we know an optimization exists.

Also I'm not worried that MDN represents any real scaling risk, except as a subset of the risks posed by our own incompetence. I know it's possible for an amateur effort to generate 50,000 pages of PDF docs in minutes (not hours) and we're not doing anything as deliberately retrograde as that. 😆

peterbe:
@ddbeck You're absolutely right. We're on the same page.

One thing that's on my mind is if you have 10,000 pages and you want to generate a whole static site (ie. 10,000 index.html files) for "every" pull request (e.g. pullrequest-1234.stumptown.cloudfront.com) is that even possible. I'm sure it is but the solution might be to have checksums on stuff so you know you only need to re-create 0.01% of pages if only some of the content is changing. 

This isn't just about memoizing little JS functions but about bigger smarter ideas. Either way, it's definitely a user story :)

Another interesting user story, to me, is some `watchman` script that watches the files in stumptown-content and runs the stumptown-renderer cli. So when you're editing `.../video/docs.md` in my editor and hit Cmd-S, it should immediately (using hot reloading) refresh `http://localhost:3000/docs/Web/HTML/Elements/video/` so I can see my edits in the browser.

Issue #93, opened 2019-08-02
Rename markdown files?


ddbeck:
Right now, most of the markdown files have one name: `docs.md`. This makes it confusing to have multiple markdown files open in an editor at once. It'd be nice to have more descriptive names (for example, `html/reference/elements/abbr/abbr.md` instead of `html/reference/elements/abbr/docs.md`).

To be able to do this, we need to figure out the rules for deciding which files to build from. Previously, `meta.yml` file was the giveaway that a folder contained the files needed to build JSON. @wbamberg raised some possible solutions (and noted some drawbacks) in https://github.com/mdn/stumptown-content/pull/91#issuecomment-517350176; there may be others.

We ought to:

- [ ] Decide on a naming convention and a method for selecting files to build from
- [ ] Merge a PR that changes the build and renames the files

ddbeck:
One half-baked idea I've had: can we use something in the front matter of a file as a way to identify the root markdown file of a page? For example, can we walk the directory tree looking for markdown files with a `recipe` front matter field as a way to identify buildable units of content?

wbamberg:
I've been thinking about guide pages today, and I think we will need to fix this problem for guide pages, where you really want something like "/html/guides" containing a pile of sensibly-named MD files, one for each guide. 

> can we walk the directory tree looking for markdown files with a recipe front matter field as a way to identify buildable units of content?

This sounds promising and I will have a go at doing this for a first-version implementation of guide pages. But I do also like the last option [here](https://github.com/mdn/stumptown-content/pull/91#issuecomment-517350176), of driving the collection-of-pages step from an explicit listing, rather that scanning everything under./content.

ddbeck:
Maybe we can combine strategies? For instance, we could use a manifest for guides and other loosely-structured content and use a scanning strategy for references. Or allow manifest files to collect whole directories with a wild card.

The reason I suggest this is that when I set a team up with Sphinx, the `toctree` directives were a big source of writer discontent. Some of the writers took to it immediately, others wouldn't stop telling me how much it frustrated them (they thought adding a file to a relevant directory was a sufficiently explicit declaration that they wanted to include that content in the output).

Alternatively, we can scan everything in `/content` as a linting strategy: if something is in `/content` but not in a manifest, then you get a warning or error. Sphinx can do this and it's OK, but it does lead to lots of annoyed "ugh update ToC" commits.

wbamberg:
[...]

PR #100, opened 2019-08-09
Add a guide type

wbamberg:
Here's a PR to add support for guides.

The thing about guides is that they are unstructured. They're prose, with some special bits (at least live samples and BCD tables, and I have also seen interactive examples occasionally) which can occur just anywhere. So this models them like KS macros: the author can embed these special bits in the Markdown using some special syntax.

The actual syntax used in this PR is like:

    {{<directive-name>:<directive-params>}}

e.g.

    {{embed-example:path/to/example}}

or

    {{embed-bcd:path.to.bcd}}

The way that each special bit (e.g. a live sample) is represented and processed is the same as for the structured content. At the moment the only special bit supported is examples, but it would be simple to add BCD and interactive-examples.

The JSON output is kind of like the output @ddbeck suggested in https://github.com/mdn/stumptown-content/issues/68#issuecomment-516347924, where we get an array of pieces of the page, where a piece might be a chunk of prose or it might be some special thing like a live sample:

```
{
  "title": "Applying color to HTML elements using CSS",
  "mdn_url": "https://developer.mozilla.org/en-US/docs/Web/HTML/Applying_color",
  "related_content": [
  ],
  "content": [
    {
      "type": "prose",
      "content": "<p>The use of color is a fundamental form of human expression..."
    },
    {
      "type": "example",
      "content": {}
    },
    {
      "type": "prose",
      "content": "<p>the next bit..."
    },
    {
      "type": "example",
      "content": {}
    },
    {
      "type": "prose",
      "content": "<p>the last bit..."
    }
  ]
}
```

I've written some code to parse these things, and included an HTML guide page (https://developer.mozilla.org/en-US/docs/Web/HTML/Applying_color) to show what it would look like.

Aside from details like how to represent these not-macros and how to parse it more sensibly than I am doing, there are a few things to think about here.

******

First, files. For this I've just dumped the MD files directly under html/guides, with sensible names ("Applying_color.md"). This is in contrast to the reference docs, where each item gets its own directory.

This means we have an inconsistency between guide pages and reference pages. Because each reference page gets its own directory, it also can have its own "examples" and "attributes" directories. But this guide page doesn't get its own "examples" directory: instead there's a communal one which it shares with all the other guide pages. I don't think we could do this for HTML elements (say) because they do need a way to distinguish (say) the `width` attribute for `canvas` and the `width` attribute for `iframe`, and at the moment that is handled simply by each element having its own "attributes" directory. I also like the idea that all the doc pieces for the `canvas` page live together: I think this helps authoring.

So in general I like the way guide pages don't get their own directories, and I like the way reference pages do get their own directories. I just don't know how I feel about the inconsistency.

One concrete problem is "contributors.md", which the guide pages need, but which can't have the same name in this model. I'm not sure what's best to do about this yet, but: I don't want this particular issue to shape the authoring model. The "contributors.md" file is effectively read-only legacy info, so it doesn't really need to be exposed to authors at all.

******

Second, the original page makes quite heavy use of `<dt>` elements, which the converter here seems to turn into `<ul>`. This is going to be a persistent issue. I guess the options are:

1. convince everyone that `<ul>` is fine
2. use (or write) an extension to remark to support them
3. use HTML in these cases

I suspect that 2 is the way we will end up going.

******

Third, the original page is a great example of a place where live samples will be "different" in stumptown. In Kuma, you identify a section of the page that contains the live sample sources, and all code blocks in that section get included in the example. This enables the author to sprinkle text among bits of code at will, and also makes the example much more integrated into the page (it just looks like page content, and some code blocks, then a magic iframe). Here's the example on MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Applying_color#Specifying_colors_in_stylesheets.

In stumptown:

1. we only allow a single description block for the whole example
2. because the example is authored independently of the article it's embedded in, we're not expecting it would be integrated seamlessly into the article. I think it would look more like an interactive example or an embedded CodePen instance, where it's clearly its own thing adjacent to the main article.

We could address both of these at the cost of more complexity. But I'm not sure it's worth it. In particular I think I like making the example more obviously an example, rather than some page content with a magic output iframe. And personally I don't like the model where code is explained line by line, and the code is very disarticulated. But I could see this being arguable.

******

Finally. There are some things that wanted to come along with the PR, that I have not included yet but would like to.

As noted above, the output here is a lot like the output @ddbeck suggested in https://github.com/mdn/stumptown-content/issues/68#issuecomment-516347924. Now I've tried this for guide pages, I'm more tempted to do it for reference pages, for consistency. I'm still worried that this option is hostile to non-MDN consumers of the JSON though.

This is the first time the code in build-json is wanting to build more than one thing, and it still hardcodes a bunch of stuff about HTML elements. I would like to try to make this code generic, so it actually uses the recipe to build the JSON.

And as also mentioned above, this now uses https://github.com/mdn/stumptown-content/issues/93#issuecomment-517636421 to identify things to build. So there would be nothing to stop us giving HTML element files sensible names.

### DIFF
--- a/content/html/guides/Applying_color.md
+++ b/content/html/guides/Applying_color.md
@@ -22,47 +22,47 @@ At a fundamental level, the [`color`](/en-US/docs/Web/CSS/color "The color CSS p
 
 Whenever an element is rendered, these properties are used to determine the color of the text, its background, and any decorations on the text.
 
--   [`color`](/en-US/docs/Web/CSS/color "The color CSS property sets the foreground color value of an element's text and text decorations, and sets the currentcolor value.")
+- [`color`](/en-US/docs/Web/CSS/color "The color CSS property sets the foreground color value of an element's text and text decorations, and sets the currentcolor value.")
 
     The color to use when drawing the text and any [text decorations](/en-US/docs/Learn/CSS/Styling_text/Fundamentals#Font_style_font_weight_text_transform_and_text_decoration) (such as the addition of under- or overlines, strike-through lines, and so forth.
 
--   [`background-color`](/en-US/docs/Web/CSS/background-color "The background-color CSS property sets the background color of an element.")
+- [`background-color`](/en-US/docs/Web/CSS/background-color "The background-color CSS property sets the background color of an element.")
 
     The text's background color.
 
--   [`text-shadow`](/en-US/docs/Web/CSS/text-shadow "The text-shadow CSS property adds shadows to text. It accepts a comma-separated list of shadows to be applied to the text and any of its decorations.")
+- [`text-shadow`](/en-US/docs/Web/CSS/text-shadow "The text-shadow CSS property adds shadows to text. It accepts a comma-separated list of shadows to be applied to the text and any of its decorations.")
 
     Configures a shadow effect to apply to text. Among the options for the shadow is the shadow's base color (which is then blurred and blended with the background based on the other parameters). See [Text drop shadows](/en-US/docs/Learn/CSS/Styling_text/Fundamentals#Text_drop_shadows "We hoped you enjoyed playing with text in this article! The next article will give you all you need to know about styling HTML lists.") in [Fundamental text and font styling](/en-US/docs/Learn/CSS/Styling_text/Fundamentals "We hoped you enjoyed playing with text in this article! The next article will give you all you need to know about styling HTML lists.") to learn more.
 
--   [`text-decoration-color`](/en-US/docs/Web/CSS/text-decoration-color "The text-decoration-color CSS property sets the color of decorations added to text by text-decoration-line.")
+- [`text-decoration-color`](/en-US/docs/Web/CSS/text-decoration-color "The text-decoration-color CSS property sets the color of decorations added to text by text-decoration-line.")
 
     By default, text decorations (such as underlines, strikethroughs, etc) use the `color` property as their colors. However, you can override that behavior and use a different color for them with the `text-decoration-color` property.
 
--   [`text-emphasis-color`](/en-US/docs/Web/CSS/text-emphasis-color "The text-emphasis-color CSS property sets the color of emphasis marks. This value can also be set using the text-emphasis shorthand.")
+- [`text-emphasis-color`](/en-US/docs/Web/CSS/text-emphasis-color "The text-emphasis-color CSS property sets the color of emphasis marks. This value can also be set using the text-emphasis shorthand.")
 
     The color to use when drawing emphasis symbols adjacent to each character in the text. This is used primarily when drawing text for East Asian languages.
 
--   [`caret-color`](/en-US/docs/Web/CSS/caret-color "The caret-color CSS property sets the color of the insertion caret, the visible marker where the next character typed will be inserted.")
+- [`caret-color`](/en-US/docs/Web/CSS/caret-color "The caret-color CSS property sets the color of the insertion caret, the visible marker where the next character typed will be inserted.")
 
-    The color to use when drawing the [caret](/en-US/docs/Glossary/caret 'caret: A caret (sometimes called a "text cursor") is an indicator displayed on the screen to indicate where text input will be inserted.') (sometimes referred to as the text input cursor) within the element. This is only useful in elements that are editable, such as [`<input>`](/en-US/docs/Web/HTML/Element/input "The HTML &lt;input> element is used to create interactive controls for web-based forms in order to accept data from the user; a wide variety of types of input data and control widgets are available, depending on the device and user agent.") and [`<textarea>`](/en-US/docs/Web/HTML/Element/textarea "The HTML &lt;textarea> element represents a multi-line plain-text editing control, useful when you want to allow users to enter a sizeable amount of free-form text, for example a comment on a review or feedback form.") or elements whose HTML `contenteditable` attribute is set.
+    The color to use when drawing the [caret](/en-US/docs/Glossary/caret) (sometimes referred to as the text input cursor) within the element. This is only useful in elements that are editable, such as [`<input>`](/en-US/docs/Web/HTML/Element/input "The HTML &lt;input> element is used to create interactive controls for web-based forms in order to accept data from the user; a wide variety of types of input data and control widgets are available, depending on the device and user agent.") and [`<textarea>`](/en-US/docs/Web/HTML/Element/textarea "The HTML &lt;textarea> element represents a multi-line plain-text editing control, useful when you want to allow users to enter a sizeable amount of free-form text, for example a comment on a review or feedback form.") or elements whose HTML `contenteditable` attribute is set.
 
 ### Boxes
 
 Every element is a box with some sort of content, and has a background and a border in addition to whatever contents the box may have.
 
--   [Borders](#Borders)
+- [Borders](#Borders)
 
     See the section [Borders](#Borders) for a list of the CSS properties you can use to set the colors of a box's borders.
 
--   [`background-color`](/en-US/docs/Web/CSS/background-color "The background-color CSS property sets the background color of an element.")
+- [`background-color`](/en-US/docs/Web/CSS/background-color "The background-color CSS property sets the background color of an element.")
 
     The background color to use in areas of the element that have no foreground content.
 
--   [`column-rule-color`](/en-US/docs/Web/CSS/column-rule-color "The column-rule-color CSS property sets the color of the rule (line) drawn between columns in a multi-column layout.")
+- [`column-rule-color`](/en-US/docs/Web/CSS/column-rule-color "The column-rule-color CSS property sets the color of the rule (line) drawn between columns in a multi-column layout.")
 
     The color to use when drawing the line separating columns of text.
 
--   [`outline-color`](/en-US/docs/Web/CSS/outline-color "The outline-color CSS property sets the color of an element's outline.")
+- [`outline-color`](/en-US/docs/Web/CSS/outline-color "The outline-color CSS property sets the color of an element's outline.")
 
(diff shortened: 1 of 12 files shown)


ddbeck:
Hey Will, thanks for opening this PR. There's a lot of ground to cover here and I'm not sure where to begin (except one point, below). What are you looking for here in terms of feedback?

But about that custom syntax: would you be open to moving the custom syntax stuff to another PR or issue, maybe replacing those KumaScript-style macro calls here with placeholder comments? I've got a bunch of questions about the syntax, the semantics implied by the implementation, and the implementation itself, but they're not really essential to guide content in particular. I'm concerned that it would overwhelm the discussion, which is why I suggest breaking into a separate PR or issue.

wbamberg:
Well let's say:

* is this general ("directive-based") approach to authoring this kind of content a good one (independent of the particular syntax used)? If not, what alternatives do we have?

* in terms of the file and directory structure, is it OK to have an inconsistency with reference pages?

* what should we do about `<dl>` non-support?

* is the way we're supporting live samples going to be too restrictive for guides? (I think it might be easier to talk about this once we have nicely rendered pages from this to look at)

* is this PR good enough for us to merge, so we can add rendering support and move closer towards having all the HTML docs in stumptown, or are there things we need to resolve before we can do that?

wbamberg:
> But about that custom syntax: would you be open to moving the custom syntax stuff to another PR or issue, maybe replacing those KumaScript-style macro calls here with placeholder comments? 

If possible I'd like this PR to deliver a piece of working functionality, and not making some decision about syntax seems to prevent that. I means this PR would be nonfunctional and would be dependent on another PR to produce something that the renderer could use.

As we discussed in email: I didn't think too much about this syntax at this point, I just wanted something easy to parse, and hoped it would provoke some better suggestions. I'm not AT ALL wedded to this syntax and choosing the KS syntax specifically was at least partly a joke (I should have been clearer about this). I feel - maybe wrongly! - that choosing a syntax is relatively simple compared with dealing with the `<dl>` issue or the question of how live samples will look in the new world.

ddbeck:
> is this general ("directive-based") approach to authoring this kind of content a good one (independent of the particular syntax used)? If not, what alternatives do we have?

For block-level directives, like you've done here, I suspect the most likely alternative is a top-down approach, a bit like recipes. You'd have some manifest file that contains the sequence of Markdown sources, BCD identifiers, and example references that would make up the built content artifact (the JSON and, ultimately, a page).

The directive-based approach makes a lot of sense for authoring this kind of content because it is, in fact, tightly coupled and sequential. Directives read more obviously in context — "and now here's where the BCD table appears" — than as just another entry in a list. They're more congruent with the intent of the content. Compare this to BCD in a reference page: BCD _is_ its own, independent unit in a reference. But in a guide, you want commentary before and after the BCD; it makes sense that you'd write it that way.

(That said, I would _not_ want to see block-level directives in our more structured content. And I have a bunch of things to say about the particular syntax.)

> in terms of the file and directory structure, is it OK to have an inconsistency with reference pages?

I'd say yes. That you're using the recipe to signal which structure to use is a step toward having a bit of a slider on structure: our tooling can have different modes where we say, "you've picked the `guide` recipe; I'm going to trust you know what you're doing" to "you've picked the `api_method` recipe; there had better not be a comma out of place". We could have stops in between, where we impose more or less structure (maybe there's a subset of guides that have a stricter structure — some routine sequence of intro, concept, BCD, list of steps, and conclusion, for example — where we use an approach more like reference pages).

> what should we do about `<dl>` non-support?

I like the idea of using conventional Markdown instead of true definition lists (`<ul>` is fine or using headings might also work). If we want true definition lists, then there's [an obvious syntax to use](https://pandoc.org/MANUAL.html#definition-lists) and a couple remark implementations floating around, or we could write our own. To me it seems sort of arbitrary, so let's assign the job of deciding to one person (draw lots?!), make them choose however arbitrarily they want to, then carry out their decision, whatever it is.

> is the way we're supporting live samples going to be too restrictive for guides? (I think it might be easier to talk about this once we have nicely rendered pages from this to look at)

I'm not bothered by it. Seems fine to let this be and revisit it if it proves to be a problem on real rendered pages.

> is this PR good enough for us to merge, so we can add rendering support and move closer towards having all the HTML docs in stumptown, or are there things we need to resolve before we can do that?

Provisionally, I'll say merge. But it would be nice if you made the embed syntax very obnoxious (I'm thinking… five braces, minimum), so we're strongly incentivized to clean it up in a follow on PR. 😄

wbamberg:
Thanks Daniel, this all seems sensible to me. I think some people might care more about `<dl>` than you.

> And I have a bunch of things to say about the particular syntax.

I'd love to hear these things, and suggestions for better syntaxes, even if we don't include them in this PR.

ddbeck:
>  I think some people might care more about `<dl>` than you.

OK, fine. Let's figure out who cares the most and make that person choose. 😆 In seriousness, I'm happy to write up the options and present the pros and cons for each, if there's someone who feels strongly enough about it to make a call.

> suggestions for better syntaxes, even if we don't include them in this PR

All right, you asked for it.

The most straightforward part of this is implementation: if we're going to add a new syntax, we should implement it as a remark plugin with its own mdast node and such. That way we'll be able to use the custom syntax in linting, spellcheck, etc., in addition to builds. This goes for whatever the syntax actually looks like.

The proposed extension syntax here looks like it can only be used for block-level items: you don't use `{{embed-bcd:path.to.bcd}}` in the middle of a sentence and the current implementation doesn't allow for it to be used inline. I'm going to assume that's intentional. Nonetheless, to me it slightly resembles some [extensions to Markdown use to set attributes with braces](https://pandoc.org/MANUAL.html#heading-identifiers), which often _can_ be used inline.

Unfortunately, GFM and CommonMark don't have a specified extension syntax yet (though [there are proposals we could draw from](https://github.com/commonmark/commonmark-spec/wiki/Generic-Directive-Extension-List)). In general, if we're going to try to do something that GFM doesn't support ordinarily, I'd like to aim for something that'll look OK rendered on GitHub and, if possible, look Markdownish (that is, looks like something that would be compatible with other popular extensions to Markdown).

In this particular use case, we don't need to be especially pretty (it will be wholly replaced, after all), we don't need to type more than a handful of times per doc, and we don't need it to work inline. I propose we (ab)use [fenced code blocks and their info string](https://github.github.com/gfm/#info-string), like this:

    ```mdn-embed
    bcd: path.to.bcd
    ```

    ```mdn-embed
    example: examples/specifying-colors-in-stylesheets
    ```

Rendered in stock GFM, it'll look like this:

```mdn-embed
bcd: path.to.bcd
```

```mdn-embed
example: examples/specifying-colors-in-stylesheets
```

But our tooling could look for code blocks with the `mdn-embed` language identifier and transform them into their own dedicated nodes for handling.

There are a number of variations on this general approach (for example, you could use `mdn` as the language and put `embed` in the code, to make the GFM-rendered form more explicit), but this is my general idea of how to handle this particular extension syntax: don't invent new (Markdown) syntax at all.

wbamberg:
>  if we're going to add a new syntax, we should implement it as a remark plugin with its own mdast node and such.

What I'm struggling with here is: in this PR the output is a JSON structure like:

```
{
  "title": "Applying color to HTML elements using CSS",
  "mdn_url": "https://developer.mozilla.org/en-US/docs/Web/HTML/Applying_color",
  "related_content": [
  ],
  "content": [
    {
      "type": "prose",
      "content": "<p>The use of color is a fundamental form of human expression..."
    },
    {
      "type": "example",
      "content": {
        "description": {
          "title": "Specifying colors in stylesheets",
          "height": 192,
          "content": "<p>The easiest way to apply color to elements..."
        },
        "sources": {
          "css": ".wrapper {\n  width: 620px;...",
          "html": "<div class=\"wrapper\">\n..."
        }
      }
    }
  ]
}
```

...like we have for reference docs. If we implement this as a Markdown plugin, are you suggesting the output of build-json (or at least the `content` property) is just a blob of HTML?

ddbeck:
I really like the output structure shown in the PR description. I'm not suggesting that you change it. I want to change how the Markdown source is consumed (and even that doesn't have to change in this PR, but I'd like an issue for it, if not).

Right now, the script uses regular expressions to slice the Markdown around the embed directives, then parses the directives and the Markdown separately. The directive never reaches the Markdown parser.

Instead, I'm suggesting that you parse the Markdown first (depending on the exact syntax we use, you may need an mdast plugin to handle directives, though if we use fenced code blocks you don't actually have to do this part, at least not initially). Then you could walk the document, accumulating top-level, non-directive nodes and their children into prose objects, but if you hit a directive node, then you can consume that node into whatever object you need (an example or BCD).

Admittedly, this all for the benefit of doing other things to this Markdown source. For example, a linter could do lightweight validity checks on our custom directives (is that BCD identifier valid? is the path to an example real?). A spellcheck could correctly skip over directives (right now `{{embed-example:path/to/example}}` is an ordinary paragraph). An editor preview pane, with a plugin, could render BCD and so on.

wbamberg:
OK, I think I understand you :).

I'm happy that the output format looks good. I'm totally happy to replace the regex approach with something better. I'll try to think about implementing what you've suggested in this PR, but if I get stuck or run out of time I'd like to merge it so we can move forward and handle your suggestion in a follow-up. It should not affect the renderer AIUI.

In preparation for that eventuality I have made the syntax more ridiculous. There are Travis errors from the MD linter: some of these were genuine errors that I have fixed, but some seem like false positive from the spellcheck, and should be fixed there. But also, we will probably eventually re-import the actual content, so I don't think there's much value in spending time massaging that at this point.

wbamberg:
OK so fairly predictably I'm gong to merge this. I have filed https://github.com/mdn/stumptown-content/issues/103 to fix the directive syntax.

Issue #108, opened 2019-08-19
Should spelling be breaking CI?


peterbe:
E.g. https://travis-ci.org/mdn/stumptown-content/builds/572764499

(if the logs disappear, here's rougly what it says)
```
> stump@0.1.0 spell-md /home/travis/build/mdn/stumptown-content
> mdspell -a -n -r -x --en-us 'content/**/!(*contributors).md'

    [1mcontent/html/guides/Applying_color.md[22m
       27 |  as the addition of under- or [31moverlines[39m, strike-through lines, and so 
       39 | orations (such as underlines, [31mstrikethroughs[39m, etc) use the `color` propert 
       97 |     Lets you draw 2D [31mbitmapped[39m graphics in a [`<canvas>`](/e 
      105 | he Web Graphics Library is an [31mOpenGL[39m ES-based API for drawing high 
      125 | s a number between 0 and 255 ([31m0x00[39m and 0xFF) or, optionally, as  
      125 | r between 0 and 255 (0x00 and [31m0xFF[39m) or, optionally, as a number  
      125 | as a number between 0 and 15 ([31m0x0[39m and 0xF). All components _mus 
      125 | ber between 0 and 15 (0x0 and [31m0xF[39m). All components _must_ be sp 
      169 | ite). Image courtesy of user [[31mSharkD[39m](http://commons.wikimedia.org 
      171 | ees (`deg`), radians (`rad`), [31mgradians[39m (`grad`), or turns (`turn`).  
      178 | 2. Then select a [31mgrayscale[39m paint that corresponds how br 
      193 | t a color. Perhaps you have a [31mcustomizable[39m user interface, or you're imp 
      215 | rk. For example, the website [[31mColorZilla[39m](http://www.colorzilla.com/)  
      226 | - [[31mPaletton[39m](http://paletton.com) 
      237 | my.org/) in association with [[31mPixar[39m](https://www.pixar.com/)) 
      239 | o express ideas. Presented by [31mPixar[39m artists and designers. 
      255 | - [[31mMedline[39m Plus: Color Blindness](https: 
      256 | - [American Academy of [31mOphthamology[39m: What Is Color Blindness?](ht 
      257 | 010/02/color-blindness.html) ([31mUsability.gov[39m: United States Department of  
      261 | there. We carefully avoid the [31mmockups[39m and the photos from movies. A 
      261 | hoto taken by one of the Mars [31mlanders[39m humanity has parked on the su 
[...]

Issue #71, opened 2019-07-12
Remove markdown-spellcheck and switch to retext


Elchi3:
I've noticed that `marked` is still in our dependency tree and I've also spend some time to further read about the package universe _unified_ that we've started to rely on. It is actually great and has tons of format and language processing packages! Really like that.

So, I'd like to propose to remove my initial spellcheck implementation (i.e. remove the `markdown-spellcheck` package) and start using `retext` and its plugins.

https://github.com/retextjs/retext is the parent package
https://github.com/remarkjs/remark-retext can get you from remark to retext

Just like `remark`, `retext` comes with many plugins, but for language processing! So, not only it does spell checking, but also it has things to do further language checking like avoiding "guys" or checking if "a" and "an" are used correctly. Wow!
https://github.com/retextjs/retext-spell
https://github.com/retextjs/retext-equality
https://github.com/retextjs/retext-indefinite-article
List of language processing plugins: https://github.com/retextjs/retext/blob/master/doc/plugins.md

I'm impressed how much validation our writing could get :)

peterbe:
> I'm impressed how much validation our writing could get :)

The bots are coming for your jobs!

...and it's both a good thing and a lie.

PR #131, opened 2019-09-12
Fix spelling warnings

wbamberg:
We currently have CI failing because of spelling errors, most of which are false positives.

We are intending to remove this particular spellcheck implementation (https://github.com/mdn/stumptown-content/issues/71) and are also discussing whether these errors should not break the build (https://github.com/mdn/stumptown-content/issues/108) but in the meantime I'd like us to have passing builds, so this is an attempt to fix the errors.


peterbe:
Thanks for making the mentions of the future work on this topic!

.spelling
```diff
@@ -68,3 +68,23 @@ Pterothorax
 resized
 encodings
 autofill
+overlines
+strikethroughs
+bitmapped
+OpenGL
+0x00
```
Elchi3: I would have preferred to put these in code like `0x00` and then the spell checker should ignore them.

Issue #101, opened 2019-08-13
Support index/landing pages


wbamberg:
MDN has "landing pages". From the [meta-docs](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Structures/Page_types#Landing_page):

> A landing page serves as a menu, of sorts, for its subpages, and is therefore primarily a navigation page. A landing page layout is typically used for the root page of a tree of pages about a particular topic. It opens with a brief summary of the topic, then presents a structured list of links to its subpages, and optionally, additional material that be useful to the reader... The list of subpages can be generated automatically...

So typically they are made of some prose content on some high-level topic like HTML, followed by a collection of links to the pages comprising the docs for this topic. Stumptown is going to need to support these things.

One basic decision is whether they are hand-maintained or (partially) generated automatically. Our actual landing pages are sometimes hand-maintained (e.g. [Learn](https://developer.mozilla.org/en-US/docs/Learn), [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript)). If landing pages are hand-maintained, then they are just guide pages and there is nothing special to do here.

But it seems quite desirable to generate them, and in that case they would need some kind of special treatment in stumptown. It would be nice if we could reuse the `related_content` thing for this - there's a sense in which landing pages are just like sidebars except they get to occupy the whole page, and it would be good for consistency if the landing page and sidebars contain the same basic content, in the same organization.

Landing pages might have at least two things that sidebars don't:

* an overview prose section
* a short description of each linked page, so readers get more than just the link text to explain what the linked page is about.

I wonder if both of these things could be just extra properties of the `related_content` object. Then the renderer could omit them when it's rendering `related_content` as a sidebar, and include them when it is rendering `related_content` as a landing page.

One thing here is: if `related_content` contains short descriptions for pages, then it seems to make sense to use the `short_description` property for that. But guide pages as currently specified don't have `short_description`...

*****

**Acceptance criteria:**
* structure for all needed HTML landing pages are defined, written and checked into stumptown-content
* code to process them into a form usable by stumptown-renderer is written and checked into stumptown-content

wbamberg:
[...]

PR #134, opened 2019-09-16
Landing pages

wbamberg:
This PR implements a page type for landing pages, along the lines suggested by https://github.com/mdn/stumptown-content/issues/101#issuecomment-530500755.

## What this PR contains
* some code to turn "landing-page" MD files into JSON that the renderer can use (build-landing-page-json.js)
* some refactoring of the surrounding code (in particular because the primitives you can use to build landing pages (`chapter_list`, `directory`) are reused from the `related_content` stuff used to build sidebars)
* a couple of sample landing pages, *content/html/HTML.md* and *content/html/reference/elements.md*. These are supposed to be for https://developer.mozilla.org/en-US/docs/Web/HTML and https://developer.mozilla.org/en-US/docs/Web/HTML/Element. They are not finished, because not all the pages they will want to link to are migrated yet. But they are there to show an idea of how an author could use this.
* some more *learn/html* pages, just to give the landing pages more stuff to link. The actual content here is just placeholder at the moment.

## What it supports
With this PR you specify a landing page as discussed in https://github.com/mdn/stumptown-content/issues/101#issuecomment-530500755, "As a special structured type". Basically a landing page here is just an overview followed by an array of "link lists" (possibly confusing name). Each link list is...a list of links, consisting of `title`, `mdn_url`, and optionally `short_description`.

You can specify a link list as (1) a literal list of links, (2) a `chapter_list` (which just enables you to reuse a list of links), (3) a `directory` (which enables you to include all pages in that directory). However you specify it, these things get built into the same JSON structure.

It includes short descriptions in the JSON if the target pages themselves contain one, and otherwise just omit them. This means we could land this PR without having to add `short_description` support into guide pages, although I do think we will want to do that anyway eventually.

The current implementation of short descriptions here is horrible for performance, because we slice the Markdown for them. It would be much better to be able to use the built JSON to get short description, but that I think would get a bit complicated. We should do this eventually but I'm not sure it's needed for the experiment.


scripts/build-json/build-landing-page-json.js
```diff
@@ -0,0 +1,59 @@
...
+        title: data.title,
+        mdn_url: data.mdn_url,
+        related_content: await related.buildRelatedContent(data.related_content),
+        body: [ overview, ...linkLists]
+    };
+}
+
+module.exports = {
+    buildLandingPageJSON
+}
```
peterbe: Sorry for being a pain but I can see myself making a PR on `stumptown-content` give it the same `prettier` config that [we have in `stumptown-renderer`](https://github.com/mdn/stumptown-renderer/commit/29672d0d97951e8449c36b8f2b267ddbfebd5237). If you just make this 2-space indent that PR won't bulldozer over this file.

peterbe: Oh. I see now. Some files are 2-space. Some are 4-space. Let's just face the inevitable and make that switch some day. Ignore me for now.

content/html/HTML.md
```diff
@@ -0,0 +1,23 @@
...
+
+An HTML element is set off from other text in a document by "tags", which consist of the element name surrounded by "`<`" and "`>`". The name of an element inside a tag is case insensitive. That is, it can be written in uppercase, lowercase, or a mixture. For example, the `<title>` tag can be written as `<Title>`, `<TITLE>`, or in any other way.
+
+The articles below can help you learn more about HTML.
```
peterbe: Maybe I've been staring myself blind and now my brains fried, but the JSON output of this .md file is this:
```json
  "body": [
    {
      "type": "prose",
      "value": {
        "title": "Overview",
        "id": "overview",
        "content": "<p><strong>HTML</strong> (HyperText Markup Language) is the most bas....snip..."
      }
    },
    {
      "type": "link_list",
      "value": {
        "title": "Learn HTML",
        "content": [
          {
            "title": "Introduction to HTML",
            "mdn_url": "https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML"
          },
          {
            "title": "Multimedia and Embedding",
            "mdn_url": "https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding"
          }
        ]
      }
    },
    {
      "type": "link_list",
      "value": {
        "title": "Reference",
        "content": [
          {
            "title": "HTML elements",
            "mdn_url": "https://developer.mozilla.org/en-US/docs/HTML/Elements"
          }
        ]
      }
    }
  ]
```

Why are the link_list sections at the bottom of this? And where does it say that they should be at the bottom?

**What if my desired final web page HTML should be something like this:**

```html
<div class="content">

  <h2>HTML: Hypertext Markup Language</h2>
 
  <!-- from Prose component -->
  <p><b>HTML</b> (HyperText Markup Language) is the most basic building block of the Web.</p>

  <!-- from LinkList component -->
  <div class="link-list">
    <h3>Links: Learn HTML</h3>
    <ul>
      <li><a href="g/en-US/docs/Learn/HTML/Introduction_to_HTML">Introduction to HTML</a></li>
      <li><a href="/en-US/docs/Learn/HTML/Multimedia_and_embedding">Multimedia and Embedding</a></li>
    </ul>
  </div>

  <!-- from Prose component -->
  <p>So, check out those links and when you're done check out these other cool references below</p>

  <!-- from LinkList component -->
  <div class="link-list">
    <h3>Links: Reference</h3>
    <ul>
      <li><a href="/en-US/docs/HTML/Elements">HTML elements</a></li>
    </ul>
  </div>

  <!-- from Prose component -->
  <p>If you've read this far, you must be really interested in this. Please also check out our <a href="/en-US/docs/Web/CSS">CSS reference docs</a></p>


</div>
```

I'm not arguing that I want/need this but I just don't understand the rationale that link_list components always appear after all the other prose.

content/html/reference/elements.md
```diff
@@ -0,0 +1,10 @@
+---
+title: HTML elements
+mdn_url: https://developer.mozilla.org/en-US/docs/HTML/Elements
+recipe: landing-page
+related_content: /content/related_content/html.yaml
+link_lists:
+  - title: Index of HTML elements
+    directory: "/content/html/reference/elements"
```
peterbe: I'm not sure if I'm nitpicking here but this line could be:
```
pages: "/content/html/reference/elements"
```
and the `build-json-*.js` scripts could worry about noticing that this is a directory and if so it basically replaces this, almost, like a macro, by `*.md` sorted alphabetically.

content/html/HTML.md
```diff
@@ -0,0 +1,23 @@
+---
+title: "HTML: Hypertext Markup Language"
+mdn_url: https://developer.mozilla.org/en-US/docs/Web/HTML
+recipe: landing-page
+related_content: /content/related_content/html.yaml
+link_lists:
+  - title: Learn HTML
+    pages:
+      - "/content/learn/html/Introduction_to_HTML.md"
```
peterbe: I think I'd prefer either `"./Introduction_to_HTML.md"` or `"/learn/html/Introduction_to_HTML.md.md"`. 
That feels more right and is much more similar to how NodeJS works. That `content` folder is not for human consumption. Similarly, now you never type `const lib = require('node_modules/some-lib')`.

But I don't feel strongly about it. I just prefer more "real" paths.

content/html/HTML.md
wbamberg: > Why are the link_list sections at the bottom of this? And where does it say that they should be at the bottom?

Over in https://github.com/mdn/stumptown-content/issues/101#issuecomment-530500755 I've talked about three different ways of implementing landing pages:

1. as guide pages
2. as guide pages with extra directives for lists of links
3. as a special structured type

This PR implements the third of these options. The structure definition says you get an overview, then a bunch of links. If that is too restrictive and there are extra requirements for a landing page (like, being able to have text at the end, or being able to have extra text fragments in between lists of links) then we should consider how to support them.

But you'll quite quickly get to a point where it's not possible to define sensible structures for these pages, and you'll just have to go with option 2.

I do want to support legitimate requirements for authors, but I'm resistant to add complexity just because we think we might want it.

scripts/build-json/build-landing-page-json.js
wbamberg: I'd happily merge a PR for this repo that would automatically fix all my formatting inconsistencies....

content/html/reference/elements.md
wbamberg: Yes, that would work fine. I don't feel strongly either way about this. I think it's one of those places where you just choose between on explicit ("directory"/"pages") or implicit ("pages" and interpret it the right way).

content/html/HTML.md
wbamberg: I don't think "./Introduction_to_HTML.md" works because we're not in the same directory. We could do "/learn/html/Introduction_to_HTML.md".

I'd be happy to make this change but I expect it'll go a bit deeper than this because there are probably other places where we are including "content" in the path.

content/html/HTML.md
peterbe: In other words; "Here's a thing. It's not the most flexible thing in the world, but it moves us forward."
I'll take it!

Actually, this reminds us of the essence of Stumptown too: simplicity. Being able to do less isn't necessarily about lack of flexibility as much as it's a matter of easier to comprehend.

content/html/reference/elements.md
peterbe: Been thinking more about it and I'm feeling more strongly about this one but I think it can wait because it's a bit out of context. A bit like HTML docs refers to images as either 
```html
<img src="../image.png"> or <img src="/images/foo.png">
```
but they don't do...
```html
<img src="/django_root/images/bar.png">
```

Speaking of images, some day we'll have images and then it becomes important that those images actually resolve when viewed in the GitHub UI for markdown preview I think. 

Long story short; some other day.

content/html/HTML.md
peterbe: Been thinking more about it and I'm feeling more strongly about this one but I think it can wait because it's a bit out of context. A bit like HTML docs refers to images as either 
```html
<img src="../image.png"> or <img src="/images/foo.png">
```
but they don't do...
```html
<img src="/django_root/images/bar.png">
```

Speaking of images, some day we'll have images and then it becomes important that those images actually resolve when viewed in the GitHub UI for markdown preview I think. 

Long story short; some other day.

content/html/reference/elements.md
peterbe: Heck! I take back what I said. It actually reads, mentally, nicely like this.

PR #135, opened 2019-09-17
Add height to interactive_example property

wbamberg:
This adds a property to `interactive_example` to represent the desired height of the iframe hosting the rendered example. It's effectively the same as the second argument to the [EmbedInteractiveExample](https://github.com/mdn/kumascript/blob/master/macros/EmbedInteractiveExample.ejs) macro.

Currently in Kuma/interactive-examples, there are the following height identifiers and corresponding values:

* JS:
    * default (no argument): 490px
    * "shorter": 410px
    * "taller": 692px

* CSS:
    * default (no argument): 390px

* HTML:
    * "tabbed-shorter": 375px
    * "tabbed-standard": 445px
    * "tabbed-taller": 655px

So they are all different. We might choose to use the same values across all three types of example, but for now I have done what seems like the simple thing and introduced keywords "html-shorter", "html-standard" and "html-taller".

This PR will break the renderer. I'm expecting the renderer to map these onto specific values in https://github.com/mdn/stumptown-renderer/blob/master/client/src/ingredients/interactive-example.js#L13. Perhaps it would be better for the content to supply an actual value in px? But I want to emphasise that you can't just pass any old value here, you can only have short, standard, tall. Then perhaps it would be better for the author to supply "html-shorter", then stumptown-content to map that to 375, then have the renderer use that directly?

I'd like to say as well that this kind of authoring use case, where we make a similar change to multiple files, is so much nicer and safer when we use GitHub, compared to the Wiki :).


peterbe:
That reminds me, imagine that we extend our Travis so that it runs a full `make deployment-build` from `stumptown-renderer` every time, as part of the CI, it would break on a PR like this. 

I guess that's why it would be nice to have that thing compiles the stumptown-content branch with the `master` of stumptown-renderer and produces a complete, with its own domain, site and posts a PR comment if it managed to succeed.

peterbe:
Why can't it just be
```
  height: "short|standard|tall"
```
(And *omitted* would imply `height: standard`. )

Why does it need to be different from JS, CSS, and HTML reference docs?

wbamberg:
> Why does it need to be different from JS, CSS, and HTML reference docs?

Because the editors are different, and want different heights. For example, JS "standard" height is 490px. CSS height is always 390px.

If you give CSS examples 490px, they will get an ugly extra bit of gray on the bottom:

****

Screen Shot 2019-09-19 at 11 35 15 AM

*****

If you give the JS examples 390px, the editor gets cut off:

******

Screen Shot 2019-09-19 at 11 36 30 AM

******

I think it might be worth standardising, but that will take some work on the interactive-examples side, and I don't think we should standardise just for the sake of it. That is, if 390px is actually the best height for the CSS examples, we should use it even if it makes it different.

But I do feel, having made this PR, that perhaps the mapping from keyword to pixel-height should be done in stumptown-content, so the built JSON just contains a pixel-height to use for the iframe.

wbamberg:
> Ergonomically, I think calling it "html-tall" or "css-standard" or "js-short" would be best and that we maintain a little map of these somewhere where build-json can reach it.

So this means that the mapping from keyword to pixels is in stumptown-content, and the JSON contains pixel values? Let me know if you prefer that and I'll make the change (I think I prefer it that way).

peterbe:
> So this means that the mapping from keyword to pixels is in stumptown-content, and the JSON contains pixel values? Let me know if you prefer that and I'll make the change (I think I prefer it that way).

Yup. And it makes sense because only an author working on the content would really understand the needs for heights on the iframes. Go for it.

Issue #118, opened 2019-08-28
Can we flatten additional_prose?


peterbe:
Spawned from this dicussion: https://github.com/mdn/stumptown-renderer/pull/89/files#r318050706

In that PR, the renderer deals very differently with `additional_prose` than it does with regular `prose`. In the packaged JSON the data struct is something like this:

```json
{
  "body": [
    {
      "type": "prose",
      "value": {
        "title": "Short description",
        "id": "short_description",
        "content": "<p>The <strong>HTML Video element</strong> ....</p>"
      }
    },
    {
      "type": "additional_prose",
      "value": [
        {
          "type": "prose",
          "value": {
            "title": "Styling with CSS",
            "content": "<p>Use the\n<a href=\"/en-US/docs/Web/CSS/list-style-image\">..."
          }
        }
      ]
    },
  ]
}
```

Why not just flatten that and manufacture an `id` so it becomes like:

```json
{
  "body": [
    {
      "type": "prose",
      "value": {
        "title": "Short description",
        "id": "short_description",
        "content": "<p>The <strong>HTML Video element</strong> ....</p>"
      }
    },
    {
      "type": "prose",
      "value": {
            "title": "Styling with CSS",
            "id": "styling_with_css",
            "content": "<p>Use the\n<a href=\"/en-US/docs/Web/CSS/list-style-image\">..."
        }
    },
  ]
}
```

wbamberg:
In https://github.com/mdn/stumptown-content/pull/107 I wasn't sure whether to do this, but decided not to because `additional_prose` sections felt like a different kind of thing to the named prose sections. They don't have language-independent IDs in the way that `short_description` does. But yes, perhaps it would be better after all to treat them the same way.

ddbeck:
Even if we have an `additional_prose` type that's different from `prose`, I think it still makes sense to flatten them (so if there are two additional prose sections, you get two `additional_prose`-type objects, instead of an object with an array of children).

peterbe:
Flattening it would really help the renderer since then it'd be able to just loop over `document.body` straight up without needing a way to self-recurse. 
For example, I was hacking on the code that adds syntax highlighting to `document.body[x]` elements that are HTML and have `<pre><code>` tags in them.

peterbe:
By the way, I'm very familiar with how the renderer worries about these. And there, I'm pretty sure it'd be fine if flattened. 
[...]

PR #137, opened 2019-09-18
Flatten additional_prose sections

wbamberg:
Fixes https://github.com/mdn/stumptown-content/issues/118.

This treats additional prose sections just like named prose sections, except they don't have an ID. We could add an ID derived from the title if you would like.

I think it would be good for the renderer to be able to handle prose sections quite liberally though, because prose sections in guide pages won't even have a title. So perhaps the renderer could say:

* if the prose section has an ID and the ID is on the "suppress headings" list, render using `Prose`
* otherwise, if the prose section does not have a title at all, render using `Prose`
* otherwise, render using `ProseWithHeading`

### DIFF
--- a/scripts/build-json/build-page-json.js
+++ b/scripts/build-json/build-page-json.js
@@ -34,7 +34,7 @@ function writeToFile(json, elementPath) {
 async function processMetaIngredient(elementPath, ingredientName, data) {
     switch (ingredientName) {
         case 'interactive_example':
-            return data.interactive_example;
+            return data.interactive_example || null;
         case 'browser_compatibility':
             return packageBCD(data.browser_compatibility);
         case 'attributes':
@@ -72,28 +72,23 @@ async function buildFromRecipe(elementPath, data, content) {
     // for each ingredient in the recipe, process the item's ingredient
     for (let ingredient of recipe.body) {
         const [ingredientType, ingredientName] = ingredient.replace(/\?$/, '').split('.');
-        // meta ingredients
         if (ingredientType === 'meta') {
+            // non-prose ingredients, which are specified in front matter
             const value = await processMetaIngredient(elementPath, ingredientName, data);
-            if (value) {
+            if (value !== null) {
                 item.body.push({
                   type: ingredientName,
                   value: value
                 });
             }
-        // additional (unnamed) prose sections
         } else if (ingredientType === 'prose' && ingredientName === '*') {
+            // additional (unnamed) prose sections
             const additionalProse = proseSections.filter(section => !section.value.id);
-            if (additionalProse.length) {
-                item.body.push(...additionalProse);
-            }
-        // named prose sections
+            item.body.push(...additionalProse);
         } else if (ingredientType === 'prose') {
+            // named prose sections
             const matches = proseSections.filter(section => section.value.id === ingredientName);
-            if (matches.length) {
-              item.body.push(matches[0]);
-            }
-        // urecognized type
+            item.body.push(...matches);
         } else {
             throw new Error(`Unrecognized ingredient type: ${ingredientType} in ${elementPath}`);
         }
--- a/scripts/build-json/build-page-json.js
+++ b/scripts/build-json/build-page-json.js
@@ -56,52 +56,48 @@ async function processMetaIngredient(elementPath, ingredientName, data) {
     }
 }
 
-async function processProseIngredient(ingredientName, proseSections) {
-    if (ingredientName !== '*') {
-        const matches = proseSections.filter(section => section.value.id === ingredientName);
-        if (matches.length) {
-            return matches[0];
-        } else {
-            return null;
-        }
-    } else {
-        const value = proseSections.filter(section => !section.value.id);
-        return {
-          type: 'additional_prose',
-          value
-        };
-    }
-}
-
 async function buildFromRecipe(elementPath, data, content) {
-    const item = {};
-    item.title = data.title;
-    item.mdn_url = data.mdn_url;
-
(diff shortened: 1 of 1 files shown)


ddbeck:
Thanks for this PR, Will. Looks promising.

> prose sections in guide pages won't even have a title

Could you point to an example of this?

wbamberg:
> Could you point to an example of this?

For example https://github.com/mdn/stumptown-content/blob/master/content/html/guides/Applying_color.md.

In reference pages, prose sections are coherent units with an H2, like "Accessibility" or "Styling with CSS", and non-prose items occupy predefined points in between them.

But in guide pages, non-prose items (live samples and BCD tables and maybe some other things) can just be embedded anywhere in the doc, and the prose items fill in the gaps.

So a guide page could look like:

```
## Some heading

Some content, blah blah blah

{{A BCD table directive}}

Content immediately following the BCD...

## Another heading

Yet more content
```

This would get built into three items in the JSON, where the third item is like:

```
{
   type: "prose"
   value: {
        content: "<p>Content immediately following the BCD...</p><h2>Another heading..."
    }
}
```

Does that make sense?

ddbeck:
Yep, this makes sense to me! I think this helps me understand what we're doing with prose a little more deeply, even. Thanks for the extra info, Will.

scripts/build-json/build-page-json.js
```diff
@@ -57,52 +57,47 @@ async function processMetaIngredient(elementPath, ingredientName, data) {
...
+            if (additionalProse.length) {
+                item.body.push(...additionalProse);
+            }
+        // named prose sections
         } else if (ingredientType === 'prose') {
-            return await processProseIngredient(ingredientName, proseSections);
+            const matches = proseSections.filter(section => section.value.id === ingredientName);
+            if (matches.length) {
+              item.body.push(matches[0]);
+            }
```
peterbe: Both here and above, if `matches` is always an array you don't need the conditional. Just do:
```javascript
const matches = proseSections.filter(section => section.value.id === ingredientName);
item.body.push(...matches);
```

scripts/build-json/build-page-json.js
```diff
@@ -57,52 +57,47 @@ async function processMetaIngredient(elementPath, ingredientName, data) {
...
                   value: value
-                };
+                });
             }
+        // additional (unnamed) prose sections
+        } else if (ingredientType === 'prose' && ingredientName === '*') {
+            const additionalProse = proseSections.filter(section => !section.value.id);
+            if (additionalProse.length) {
+                item.body.push(...additionalProse);
+            }
```
peterbe: Same here, these 3 lines can be replaced by:
```javascript
const additionalProse = proseSections.filter(section => !section.value.id);
item.body.push(...additionalProse);
```

scripts/build-json/build-page-json.js
```diff
@@ -57,52 +57,47 @@ async function processMetaIngredient(elementPath, ingredientName, data) {
...
-    // for each ingredient in the recipe, process the item's ingredient
     const proseSections = await packageProse(content);
-    item.body = await Promise.all(recipe.body.map(async ingredient => {
+
+    // for each ingredient in the recipe, process the item's ingredient
+    for (let ingredient of recipe.body) {
         const [ingredientType, ingredientName] = ingredient.replace(/\?$/, '').split('.');
+        // meta ingredients
         if (ingredientType === 'meta') {
             const value = await processMetaIngredient(elementPath, ingredientName, data);
```
peterbe: Why does this sometimes become `undefined`? (I.e. the value of `value`). Seems odd. I'd rather `processMetaIngredient` returned `null` if there's nothing to be found. This `undefined` always makes me wonder if something went wrong.

scripts/build-json/build-page-json.js
```diff
@@ -57,52 +57,47 @@ async function processMetaIngredient(elementPath, ingredientName, data) {
...
+                item.body.push(...additionalProse);
+            }
+        // named prose sections
         } else if (ingredientType === 'prose') {
-            return await processProseIngredient(ingredientName, proseSections);
+            const matches = proseSections.filter(section => section.value.id === ingredientName);
+            if (matches.length) {
+              item.body.push(matches[0]);
+            }
+        // urecognized type
```
peterbe:
```suggestion
```
Comment is oddly placed in the preceding block and it's not adding any value since it's clear from the string used in throwing of the error below.

scripts/build-json/build-page-json.js
```diff
@@ -57,52 +57,47 @@ async function processMetaIngredient(elementPath, ingredientName, data) {
...
+    };
 
-    // for each ingredient in the recipe, process the item's ingredient
     const proseSections = await packageProse(content);
-    item.body = await Promise.all(recipe.body.map(async ingredient => {
+
+    // for each ingredient in the recipe, process the item's ingredient
+    for (let ingredient of recipe.body) {
         const [ingredientType, ingredientName] = ingredient.replace(/\?$/, '').split('.');
+        // meta ingredients
```
peterbe: Even me, I've already forgotten what this means. Would love to see a comment that explains, briefly, what a "meta" ingredient means.

scripts/build-json/build-page-json.js
```diff
@@ -34,7 +34,7 @@ function writeToFile(json, elementPath) {
 async function processMetaIngredient(elementPath, ingredientName, data) {
     switch (ingredientName) {
         case 'interactive_example':
-            return data.interactive_example;
+            return data.interactive_example || null;
```
peterbe: 👍

PR #140, opened 2019-09-19
Protect against checked in diffs

### LINKED ISSUE #139: Protect against checked in diffs
In Kuma someone fumbled a little when doing a git rebase or something. 
It accidentally [left git diff markers in the file](https://github.com/mozilla/kuma/commit/235c264d55cf5719b2505d30cbe836215ef74782#diff-77138e0db7f64114bd7795f66f0bd9ba). Had this happened to a `.py` file, it would have quickly broken the build when booting up Python because it would crash with a `SyntaxError`. But this happened to a html template (jinja) so it managed to slip by TravisCI. 

I can totally see this happening in `stumptown-content` since a Markdown conversion might not break with these diff markers in it.

peterbe:
Fixes #139

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ branches:
 
 script:
   # Make sure there are no left-over git patch markers in
-  - git --no-pager grep '>>>>>>>' content && echo "There are leftover diff markers" && exit 1
+  - npm run ci-checks
   - npm run lint-js
   - npm run lint-md
   - npm run spell-md
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ branches:
     - master
 
 script:
+  # Make sure there are no left-over git patch markers in
+  - git --no-pager grep '>>>>>>>'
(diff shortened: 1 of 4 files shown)


peterbe:
Hmm... That didn't work. It failed because of spelling, not because of the presence of the `>>>>>>>` string.

peterbe:
Spits out:
```
> node scripts/ci/git.js

The following lines look like git diff markers:
content/html/reference/elements/video/docs.md:28:>>>>>>> Please ignore me.
```

.travis.yml
```diff
@@ -6,6 +6,8 @@ branches:
     - master
 
 script:
+  # Make sure there are no left-over git patch markers in
+  - npm run ci-checks
```
wbamberg: The comment "# Make sure there are no left-over git patch markers in" looks much less generic than the name of the command. Is it the intention to add more tests into `ci-checks`, and if so will that comment age badly?

Issue #76, opened 2019-07-25
Shorter titles for the related content?


peterbe:
See screenshot in https://github.com/mdn/stumptown-renderer/pull/58

This is a very specific reference to the existing MDN pages which displays a different title for certain elements in the sidebar compared to the actual title when you get to the page. 

The stumptown-renderer is dumb and will just display the `title` key from the packaged JSON so if we want to have different titles in the sidebar compared to the document page, here's the place to resolve that. 

****

**Acceptance criteria:**
* we have a way to represent short titles in stumptown content, and the short title is included in the JSON
* the renderer uses short titles when rendering the sidebar

peterbe:
We could, in the front-matter have an optional key called "short_title". So, when we generated the `related_content` we do something like `related_content.title = doc.meta.short_title || doc.meta.title`

peterbe:
Somewhat separate to that is that on MDN the sidebar short titles are displayed with monospace font. I.e. in the sidebar it's `<code>&lt;video&gt;</code>` but when you click on it, it's `&lt;video&gt: The Video Embed element`

Here I think renderer needs to be a bit smarter so it knows to render the titles slightly differently. E.g. 
```
<li>{node.is_code ? <code>{node.title}</code> : node.title}</li>
```

But what else is there like that? ...so that we think ahead a little. What about the icons? I think it would make sense for stumptown-content to have a secret language and the in the renderer we do the visual decisions. E.g. 
```
<li className={node.is_deprecated ? 'deprecated-sidebar-item' : null}>
{node.title}
</li>
```

wbamberg:
On short titles, yes, I think we ought to do this and I think we should do basically what you say in https://github.com/mdn/stumptown-content/issues/76#issuecomment-515200733.

Also this isn't just about length, but also context. For example, recently as an SEO experiment @a2sheppy tried making some CSS pages have long titles, so rather than "margin" it said something like "CSS margin property". This didn't stick in the end, but while it was going on these titles were showing up in the sidebar, like

```
CSS
    Properties
        CSS background property
        CSS background-color property
        CSS border property
```

...it's obvious from the context that it's a CSS property, so saying it again in the link text just looks silly. It's also harder to scan this way.

So yes, it feels as if we should let pages have different titles from their names in the sidebar (or any other place where the context is already given).

a2sheppy:
> Here I think renderer needs to be a bit smarter so it knows to render the titles slightly differently. E.g.

> ```
[...]

PR #142, opened 2019-09-20
Add short_title

wbamberg:
Fix for https://github.com/mdn/stumptown-content/issues/76.

This adds a new optional item in front matter, `short_title`. If present it's added to the `related_content` entry for the item. I've included both this and `title`, since choosing which to use feels like a renderer choice.

Questions:

* should we do this? I think that if we want to have long titles for pages, like "<abbr>: The Abbreviation element" then yes, we should have a short alternative for places like the sidebar. This isn't just about the available space, it's also about context: if this link appears in a list headed "HTML elements", say, then repeating "element" for every item looks silly. But for the title of a page it may be better to supply extra context.

* should it be in the recipe? I've not included it in the recipe, since we don't include `title`. I think if we include either we should include both.

### DIFF
--- a/content/html/reference/elements/abbr/docs.md
+++ b/content/html/reference/elements/abbr/docs.md
@@ -1,5 +1,6 @@
 ---
 title: '<abbr>: The Abbreviation element'
+short_title: <abbr>
 mdn_url: https://developer.mozilla.org/docs/Web/HTML/Element/abbr
 tags:
     group: Inline text semantics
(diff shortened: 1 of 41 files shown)


peterbe:
Bah! I'm sorry. I didn't see you he had requested a review from @ddbeck as well. Sorry!
If you find anything that you would object to, we can easily revert this or just do a quick "fail forward" follow-up.

wbamberg:
> Bah! I'm sorry. I didn't see you he had requested a review from @ddbeck as well. Sorry!

No worries.

> If you find anything that you would object to, we can easily revert this or just do a quick "fail forward" follow-up.

Yes, I think that's a fine plan.

PR #143, opened 2019-09-20
Fix guide sidebar

wbamberg:
I was trying to update stumptown-renderer for the latest stumptown-content changes when I realized there was a bug: the [code to build `related_content` for guide pages](https://github.com/mdn/stumptown-content/blob/7b860a6d6c6e7ab651c6fe13e5560cbd0d51b6da/scripts/build-json/build-guide-page-json.js#L47) is missing `await`.

Then I thought, this happened because we are making the same call to build related content three times, once for landing pages, once for guide pages, and once for all the recipe-driven pages. So I tried refactoring things a bit to reduce this duplication.

There are a couple of things I don't like about it still:

* guide and landing pages specify `related_content` directly in their front matter, while recipe pages specify it in the recipe. I think there is sense to this, if all pages that share a recipe have a similar sidebar, while guide and landing pages are generic and can be anywhere in the site, so can have any sidebar they please. But it is a bit weird.

* only recipe pages support contributors. I think we ought to rethink the way we represent contributors anyway, so I have kept it as-is for now.

### DIFF
--- a/scripts/build-json/build-page-json.js
+++ b/scripts/build-json/build-page-json.js
@@ -62,15 +62,8 @@ async function processMetaIngredient(elementPath, ingredientName, data) {
     }
 }
 
-async function buildFromRecipe(elementPath, data, content) {
-    const recipePath = path.join(__dirname, '..', '..', 'recipes', `${data.recipe}.yaml`);
-    const recipe = yaml.safeLoad(fs.readFileSync(recipePath, 'utf8'));
-    const item = {
-      title: data.title,
-      mdn_url: data.mdn_url,
-      related_content: await related.buildRelatedContent(recipe.related_content),
-      body: []
-    };
+async function buildFromRecipe(elementPath, data, content, recipe) {
+    const body = [];
 
     const proseSections = await packageProse(content);
 
@@ -81,7 +74,7 @@ async function buildFromRecipe(elementPath, data, content) {
             // non-prose ingredients, which are specified in front matter
             const value = await processMetaIngredient(elementPath, ingredientName, data);
             if (value !== null) {
-                item.body.push({
+                body.push({
                   type: ingredientName,
                   value: value
                 });
@@ -89,20 +82,17 @@ async function buildFromRecipe(elementPath, data, content) {
         } else if (ingredientType === 'prose' && ingredientName === '*') {
             // additional (unnamed) prose sections
             const additionalProse = proseSections.filter(section => !section.value.id);
-            item.body.push(...additionalProse);
+            body.push(...additionalProse);
         } else if (ingredientType === 'prose') {
             // named prose sections
             const matches = proseSections.filter(section => section.value.id === ingredientName);
-            item.body.push(...matches);
+            body.push(...matches);
         } else {
             throw new Error(`Unrecognized ingredient type: ${ingredientType} in ${elementPath}`);
         }
     }
 
-    const contributorsPath = path.join(elementPath, 'contributors.md');
-    item.contributors = await packageContributors(contributorsPath);
-
-    return item;
+    return body;
 }
 
 async function buildPageJSON(docsPath) {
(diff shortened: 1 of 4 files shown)


wbamberg:
Sorry if you don't want to review all these changes, I could just make the one-liner to add `await` to https://github.com/mdn/stumptown-content/blob/7b860a6d6c6e7ab651c6fe13e5560cbd0d51b6da/scripts/build-json/build-guide-page-json.js#L47 .

scripts/build-json/build-page-json.js
```diff
@@ -116,35 +39,51 @@ async function buildPageJSON(docsPath) {
...
+                }
             default:
                 throw new Error(`Not a supported recipe: ${data.recipe}`);
         }
 
+        // build the item
+        const item = {
+          title: data.title,
+          mdn_url: data.mdn_url,
+          related_content: await related.buildRelatedContent(relatedContentSpec),
```
peterbe: Perhaps I'm just too lazy to dig in and answer my own question but... Does every page have related content?
E.g. a landing page might take up the "full screen" and not display a sidebar. Or, is it that landing pages (ab)use the related content for its internal link list building?

scripts/build-json/build-page-json.js
```diff
@@ -116,35 +39,51 @@ async function buildPageJSON(docsPath) {
...
-                item = await buildFromRecipe(docsDirectory, data, content);
-                break;
+            case 'html-element': {
+                    // for recipe-driven content, related_content is in the recipe
+                    const recipePath = path.join(__dirname, '..', '..', 'recipes', `${data.recipe}.yaml`);
+                    const recipe = yaml.safeLoad(fs.readFileSync(recipePath, 'utf8'));
+                    relatedContentSpec = recipe.related_content;
+                    body = await recipePage.buildRecipePageJSON(docsDirectory, data, content, recipe);
+                    // currently only reference-driven content supports contributors
+                    contributors = await packageContributors(path.join(docsDirectory, 'contributors.md'))
```
peterbe: Thank you for your note in the PR description. Yeah, this sucks. Whatever we do, soon, I'd like for the renderer to just be 1 (or 2) link to a page that lists contributors.

scripts/build-json/build-page-json.js
```diff
@@ -116,35 +39,51 @@ async function buildPageJSON(docsPath) {
...
                 elementDirectory = path.join(docsDirectory, path.basename(docsPath).split('.')[0]);
                 break;
-            case 'html-element':
-                item = await buildFromRecipe(docsDirectory, data, content);
-                break;
+            case 'html-element': {
+                    // for recipe-driven content, related_content is in the recipe
+                    const recipePath = path.join(__dirname, '..', '..', 'recipes', `${data.recipe}.yaml`);
+                    const recipe = yaml.safeLoad(fs.readFileSync(recipePath, 'utf8'));
+                    relatedContentSpec = recipe.related_content;
```
peterbe: I know it's going to have to change that we can't have a switch case for every single thing that isn't guide page or landing page. 
Either way, it would be nice to avoid the disk IO and "cache" the `recipe` object based on the `data.recipe` input. We can tackle that some other day.

scripts/build-json/build-page-json.js
wbamberg: I think almost every page and possibly actually every page should have a sidebar. The only one I know that doesn't is https://developer.mozilla.org/en-US/docs/Web/API, and arguably that one ought to as well, just for consistency.

scripts/build-json/build-page-json.js
wbamberg: > I know it's going to have to change that we can't have a switch case for every single thing that isn't guide page or landing page.

This has too many negatives for me to understand, but: I hope we will have just one case for all recipe-based pages. But you're right of course, we should cache the recipes here.

PR #147, opened 2019-09-24
Scrape BCD and interactive-examples front matter

wbamberg:
This PR makes the scraper scrape BCD and interactive-example front matter from the Wiki. It only works for recipe-based pages, so it also adds a new flag "--guide-page". If this flag is passed, the behavior is suppressed.

(We would want to do something different for guide pages, but since 80-90% of pages are reference pages, it's less important to make migration quicker.)

The next most important bits to scrape would be live samples, but that would be a lot harder.

### DIFF
--- a/scripts/scraper/scrape-front-matter.js
+++ b/scripts/scraper/scrape-front-matter.js
@@ -50,7 +50,7 @@ function extractBCD(html) {
 
 async function scrapeFrontMatter(title, url, md, isGuidePage) {
   const fullURL = `${baseURL}${url}`;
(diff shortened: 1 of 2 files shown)


wbamberg:
Sorry to merge this without review, but I need to start using it.

ddbeck:
@wbamberg after the fact, this looks fine.

Also, sorry about my unresponsiveness about this and other things—I got a bit overwhelmed by travel and dropped the ball on a bunch of things. Working on recovering now. 😬

wbamberg:
It's no reflection on you, Daniel, I just got fed up manually dismembering live samples, and wanted to try to scrape them, and wanted to build it on this.

Issue #123, opened 2019-09-10
Ditch mdn_url for slug + locale?


peterbe:
Here's what the `related_content` looks like:

```json
            {
              "title": "<dialog>: The Dialog element",
              "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/dialog"
            },
            {
              "title": "<div>: The Content Division element",
              "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/div"
            },
            {
              "title": "<dl>: The Description List element",
              "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/dl"
            },
            {
              "title": "<dt>: The Description Term element",
              "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/dt"
            },
            {
```
...for example. 

The domain name definitely feels redundant so that's probably a slam dunk.
But this `/docs/` thing is just a Wiki left-over that I'm not sure anybody knows anymore why we have it (but let's not worry about removing it just yet!). 
What really matters is the `Web/HTML/Element/dialog` string and when we get there it might smart to also keep the locale (e.g. `en-US`). 

On the renderer, we have a hacky function called `fixRelatedContentURIs` which, currently, just strips the `mdn_url` down to a key called `uri` which is just the path part.

Issue #75, opened 2019-07-25
related_content's mdn_url is different from the main doc


peterbe:
Looking at `packaged/html/reference/elements/abbr.json` for example.

Looks like this:
```
{
   "related_content": [
    {
      "title": "Learn HTML",
      "content": [
        {
          "title": "Introduction to HTML",
          "content": [
            {
              "title": "Introduction to HTML",
              "mdn_url": "https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML"
            },
           ...

  "title": "<abbr>: The Abbreviation element",
  "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/abbr",
  "interactive_example_url": "https://interactive-examples.mdn.mozilla.net/pages/tabbed/abbr.html",
  "browser_compatibility": {
    ...
```

Note the difference in `mdn_url`values. Some are with `en-US` and some are not. 

We've talked before about at least dropping the absolute part with the domain. No doubt. But should we go for the locale prefix? My inclination is to NOT have the locale in there. Some day, the `abbr.json` file will be called something like `packaged/en-US/html/reference/elements/abbr.json` (or `packaged/sv-SE/html/reference/elements/abbr.json`) so that piece of information will be implicit. 

However, it could be that the lazy Swedes haven't translated every page yet so they might want to link to the en-US version in their side bars because it's better to point the English content than to not link at all. So for that reason it might be good to have the locale in there.

peterbe:
By the way, the current Stumptown-renderer does NOT have locales in the URLs like MDN does. If we're ever going to have a chance to migrate, I think we should stick with the MDN-like URLs and start using en-US in stumptown-renderer for now. 

Based on that loud-thought I'm starting to think that the right thing to do is to always have the locale in the URIs. Always.

peterbe:
I think we should start encoding the locale sooner rather than later in the renderer prototype.
See https://github.com/mdn/stumptown-renderer/issues/57

We can discuss the less useful "/docs/" URI segment some other day. :)

wbamberg:
@peterbe , I think we can close this one now can't we?

peterbe:
Yup. I grepped and the only full URLs I see are in the `__compat` block. E.g.
```
          "__compat": {
            "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/ins",
```
I don't we're using that in stumptown-renderer are we now that we have the `query`?

PR #149, opened 2019-09-25
Remove locale from mdn_url

wbamberg:
Note: this is not a fix for https://github.com/mdn/stumptown-content/issues/123, in fact it's the opposite, since in https://github.com/mdn/stumptown-content/issues/123, AFAICT, we're intending to keep the locale in `mdn_url`.

However, I was trying to get guide pages rendering, and realised that the example guide page here (https://github.com/mdn/stumptown-content/blob/master/content/html/guides/Applying_color.md) isn't getting shown in stumptown-renderer's server. I *think* this is because the `mdn_url` includes the locale, so the JSON is being put in "client/.build/**en-US**/docs/Web/HTML/Applying_color/index.json". But the server is looking in "client/.build/docs/Web/HTML/Applying_color/index.json". If I remove the locale from the JSON in stumptown/packaged/, then the server does find it.

So this PR updates `mdn_url` in all places to remove the locale, and after that it should be possible to figure out how to render guide pages.

If you would prefer to change `mdn_url` instead along the lines of https://github.com/mdn/stumptown-content/issues/123, then I'm happy to do that instead. But then AFAICT we'd need corresponding updates in stumptown-renderer.

### DIFF
--- a/content/html/HTML.md
+++ b/content/html/HTML.md
@@ -1,6 +1,6 @@
 ---
 title: "HTML: Hypertext Markup Language"
-mdn_url: https://developer.mozilla.org/en-US/docs/Web/HTML
+mdn_url: https://developer.mozilla.org/docs/Web/HTML
 recipe: landing-page
 related_content: /content/related_content/html.yaml
 link_lists:
--- a/content/html/guides/Applying_color.md
+++ b/content/html/guides/Applying_color.md
@@ -1,6 +1,6 @@
 ---
 title: Applying color to HTML elements using CSS
-mdn_url: https://developer.mozilla.org/en-US/docs/Web/HTML/Applying_color
+mdn_url: https://developer.mozilla.org/docs/Web/HTML/Applying_color
 related_content: /content/related_content/html.yaml
 recipe: guide
 ---
--- a/content/html/reference/elements.md
+++ b/content/html/reference/elements.md
@@ -1,6 +1,6 @@
 ---
 title: HTML elements
-mdn_url: https://developer.mozilla.org/en-US/docs/HTML/Elements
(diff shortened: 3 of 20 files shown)


wbamberg:
This is a partial fix for https://github.com/mdn/stumptown-content/issues/75.

peterbe:
The reason http://localhost:3000/en-US/docs/Web/HTML/Applying_color isn't working is because of the naive way the react-router works in `client/src/app.js`. The reason it says "Page Not Found" is because the router doesn't know which document component to render at all. You'd get the same effect as: http://localhost:3000/xxxxxx or http://localhost:3000/zzzzzz

The reason http://localhost:3000/docs/Web/HTML/Applying_color is failing is because of the XHR request failing to find http://localhost:5000/docs/Web/HTML/Applying_color/index.json because that doesn't exist on disk. However, http://localhost:3000/en-US/docs/Web/HTML/Applying_color/index.json exists on disk. 

This is why I think we should go 100% en-US on ALL URLs everywhere. I.e. in the `mdn_url` fields in the JSON. And in the react routing and whatever else needs to be done to the `cli`. 

Long live `en-US`! (that tasted bad to write)

peterbe:
So, I prose we close this and focus on 2 things:
1) Put 'en-US' everywhere where it isn't already. 
2) refresh the React routing (I can do that, somewhere [here](https://github.com/mdn/stumptown-renderer/blob/257000fae9f924f451dc19e9bc22f23a77136310/client/src/app.js#L18))

wbamberg:
OK, sounds good. CLosed in favour of https://github.com/mdn/stumptown-content/pull/152.

PR #152, opened 2019-09-25
Updated mdn_url

wbamberg:
This is (possibly) a fix for https://github.com/mdn/stumptown-content/issues/123.

It changes `mdn_url` everywhere to be of the form:

    /en-US/docs/path/to/page

Note that this PR will break the renderer.

### DIFF
--- a/content/html/HTML.md
+++ b/content/html/HTML.md
@@ -1,6 +1,6 @@
 ---
 title: "HTML: Hypertext Markup Language"
-mdn_url: https://developer.mozilla.org/en-US/docs/Web/HTML
+mdn_url: /en-US/en-US/docs/Web/HTML
 recipe: landing-page
 related_content: /content/related_content/html.yaml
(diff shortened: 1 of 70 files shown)


peterbe:
What about the BCD data?
E.g.
```
    {
      "type": "browser_compatibility",
      "value": {
        "__compat": {
          "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/aside",
  ...
```

wbamberg:
That would be a conversation in https://github.com/mdn/browser-compat-data. I suspect that because this is used by external applications to link to our docs, they would want a complete URL.

peterbe:
Somehow someday if we ever switch from kuma to stumptown-renderer we'd need to learn to cope with a URL like `/docs/Web/HTML/aside` because there'll be no `<root>/build/docs/Web/HTML/aside/index.html` file on disk.

Issue #141, opened 2019-09-19
Node version minimum?


peterbe:
We're currently using [`node_js: stable`](https://github.com/mdn/stumptown-content/blob/093a4671553db9f53b5611298feff143ed9fd303/.travis.yml#L2) which means, as of today, version `v12.10.0`.

Should we stick to Node v10 to make sure we're not using features of Node that not everyone might have? I'd hate it if someone is using an OS that makes it really hard to upgrade to v12 and now they can't contribute. Hopefully, we're not using any "bleedy edge" features. I also don't know what the difference is between v10 and v12 or how inconvenient it might be to get version 12 installed.

ddbeck:
I don't know how to specify this in the Travis CI config, but I like the idea of supporting the current active LTS release or newer. Seems like a good balance between getting recent features and the likelihood that contributors will have access to a supported version.

peterbe:
I think you just put `node_js: 10`. [For example](https://github.com/mozilla/kuma/blob/ff52aea11226006c856084deea63ff21c9212de9/.travis.yml#L47)
However, it's my personal goal to move this repo (at least to try) to GitHub Actions so we can worry about it then.

peterbe:
Now, `node_js: stable` means **v13.1.0** in TravisCI. 
What I'd rather go for is to say that our minimum is "Stable" as Nodejs.org defines it. Or perhaps the language is to say "The latest LTS release" . 
I.e. not saying a specific number but just keep up with what LTS is. 

So that means scripts would be able to use Node v12 features (hi `String.prototype.matchAll`!) and we should change TravisCI to use that instead.

ddbeck:
Having played with GitHub Actions for BCD a little bit, an idea: an Action that checks https://github.com/nodejs/Release/blob/master/schedule.json for the most recent actively maintained LTS release. You could use it to change the travis CI config or use it a variable for the next build step. (There are some annoyances with node_modules in Actions though, so it might be worth setting up a separate repo for it.)

PR #155, opened 2019-10-02
Extend scraper to scrape live samples

wbamberg:
This PR extends the scraper to scrape live samples (that is, examples embedded using the `EmbedLiveSample` macro).

Most of the time there will be some manual follow-up work after scraping them using this tool, but it should get you 80% of the way there at least.

It finds `EmbedLiveSample` calls and works out which code blocks each one contains.

It removes these code blocks from the source doc and creates them as separate "example.*" files under an /examples/${example_id}/ directory. It sweeps up everything else it found in the example into the description.md file, and removes that from the source too.

It also writes an `examples` front matter property pointing at these directories.


scripts/scraper/process-macros/extract-macro-calls.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Given the name of a macro and some HTML,
+ * find the all invocations of the macro in the HTML.
+ *
+ * Return an array, containing one element per invocation,
+ * where each element is itself an array containing the arguments
+ * with which the macro was called in that invocation.
+ */
+function extractMacroCalls(macroName, html) {
+  const regex = new RegExp(`{{\\ ?${macroName}\\((.*?)\\)}}`, 'gi');
```
peterbe: Don't ask me why I noticed this but if this is the input:
```html
<p>{{EmbedLiveSample("Examples")}}</p>

<div>{{ EmbedLiveSample("Examples2")}}</div>

<pre>{{EmbedLiveSample('Examples3') }}</pre>
```
...it would only return `[ [ 'Examples' ], [ 'Examples2' ] ]`

peterbe: By the way, the `String.matchAll` is neat. Try
```javascript
function extractMacroCalls(macroName, html) {
    const regex = new RegExp(`{{\\s*${macroName}\\(["'](.*)["']\\)\\s*}}`, "g");
    return [...html.matchAll(regex)].map(m => m[1]);
}
```

wbamberg: `matchAll` would be great here but [it's only supported from Node 12](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#Browser_compatibility). I understood from https://github.com/mdn/stumptown-content/issues/141 that we want to support the current LTS release (which seems sensible), and that's v10.

wbamberg: > if this is the input...it would only return [ [ 'Examples' ], [ 'Examples2' ] ]

Ugh. I've pushed an update that seems to fix this particular regex bug.

peterbe: Good catch. I didn't know this so now I wonder where else I've foolishly added that.

peterbe: Mind you, my little suggestion had two other tricks that might be worth to keep. I.e. `\\s*` instead of `\\ ?` so it can match 0 or many. And the `["'](.*)["']` should mean you don't need to clean up the match in the iterator. But perhaps I've misunderstood something about how these macros always work.

wbamberg: I'm always happy to accept help with regexes as my understanding of them is a bit primitive. But this:

```js
function extractMacroCalls(macroName, html) {
    const regex = new RegExp(`{{\\s*${macroName}\\(["'](.*)["']\\)\\s*}}`, "g");
    return [...html.matchAll(regex)].map(m => m[1]);
}
```

does not seem to work, even with `matchAll` support:

```
> var s = `<p>{{EmbedLiveSample("Examples")}}</p><div>{{ EmbedLiveSample("Examples2")}}</div><pre>{{EmbedLiveSample('Examples3') }}</pre>`
> extractMacroCalls('EmbedLiveSample', s)
< Array [ "Examples\")}}</p><div>{{ EmbedLiveSample(\"Examples2\")}}</div><pre>{{EmbedLiveSample('Examples3" ]
```

scripts/scraper/process-macros/extract-macro-calls.js
```diff
@@ -0,0 +1,24 @@
...
+ * Return an array, containing one element per invocation,
+ * where each element is itself an array containing the arguments
+ * with which the macro was called in that invocation.
+ */
+function extractMacroCalls(macroName, html) {
+  const regex = new RegExp(`{{\\ ?${macroName}\\((.*?)\\) ?}}`, 'gi');
+  let allInvocations = [];
+  let thisInvocation = [];
+
+  while ((thisInvocation = regex.exec(html)) !== null) {
```
ddbeck: This is sort of cryptic. I had to re-read how `exec` works to understand that this loop will eventually exhaust the matches. I don't have any idea how to make this less cryptic, so I'd appreciate a comment here to give a hint about what's going on

wbamberg: Yes, I agree this is cryptic. I just copied the code out of MDN :).

PR #156, opened 2019-10-02
mdn-sitemap-compare

peterbe:


peterbe:
Not my best work but it works. It might be worth running CI since it'll catch some typos at least.

scripts/linting/mdn-sitemap-compare.js
```diff
@@ -0,0 +1,130 @@
...
+async function fetchSitemap(url) {
+    if (url.startsWith("/")) url = `${baseURL}${url}`;
+    let r = await fetch(url);
+    let payload = await r.text();
+    return payload;
+}
+
+async function getLocs(url) {
+    const payload = await fetchSitemap(url);
+    return [...payload.matchAll(/<loc>(.*?)<\/loc>/g)].map(m => m[1]);
```
wbamberg: As we said, `matchAll` needs Node 12. But, according to https://nodejs.org/en/about/releases/ , Node 12 will be the LTS in exactly two weeks! So maybe it's OK :).

scripts/linting/mdn-sitemap-compare.js
```diff
@@ -0,0 +1,130 @@
...
+const readFile = util.promisify(fs.readFile);
+
+const baseURL = "https://developer.mozilla.org";
+
+const ROOT = path.join(__dirname, "..", "..");
+
+async function fetchSitemap(url) {
+    if (url.startsWith("/")) url = `${baseURL}${url}`;
+    let r = await fetch(url);
+    let payload = await r.text();
```
wbamberg: I don't know how much you care about nitpicky comments, or what our policy is on this for that matter, but this and the previous line could be `const`.

scripts/linting/mdn-sitemap-compare.js
```diff
@@ -0,0 +1,130 @@
...
+}
+
+async function getLocs(url) {
+    const payload = await fetchSitemap(url);
+    return [...payload.matchAll(/<loc>(.*?)<\/loc>/g)].map(m => m[1]);
+}
+
+async function getMDNURL(filepath) {
+    const content = await readFile(filepath, "utf8");
+    const match = content.match(/mdn_url:\s*(.*)/);
```
wbamberg: I think this would be safer and clearer if it explicitly loaded the front matter:

```js
const matter = require('gray-matter');
const { data, content } = matter(docs);
```

scripts/linting/mdn-sitemap-compare.js
```diff
@@ -0,0 +1,130 @@
...
+        let has = 0;
+        const maxLength = 100;
+        ourURIs.forEach(uri => {
+            const wspace = " ".repeat(maxLength - 1 - uri.length);
+            let icon = "✅";
+            if (!allPossible[uri]) {
+                fails++;
+                icon = "🚫";
+            } else {
+                has++;
```
wbamberg: I'm not sure why we are keeping track of both of these, since `has` is just `ourURIs.length - fails`, isn't it?

scripts/linting/mdn-sitemap-compare.js
```diff
@@ -0,0 +1,130 @@
...
+            .filter(dirent => dirent.isDirectory())
+            .map(dirent => path.join(directory, dirent.name))
+            .map(findFiles)
+            .forEach(files => found.push(...files));
+
+        return found;
+    }
+    const files = findFiles(path.join(ROOT, "content"));
+    return Promise.all(files.map(getMDNURL)).then(values => {
+        const ourURIs = values.filter(v => v);
```
wbamberg: nitpicky, but `stumptownURIs` would be clearer than `ourURIs`.

scripts/linting/mdn-sitemap-compare.js
```diff
@@ -0,0 +1,130 @@
...
+            url_.includes("/sitemap.xml") &&
+            (!args.length ||
+                args.includes(url.parse(url_).pathname.split("/")[2]))
+        );
+    });
+
+    try {
+        const values = await Promise.all(urls.map(getLocs));
+        const allPossible = {};
+        values.forEach(urls => {
```
wbamberg: It's quite confusing (to me) to reuse the `urls` name, declared above as a `const` variable, for a different thing.

scripts/linting/mdn-sitemap-compare.js
```diff
@@ -0,0 +1,130 @@
...
+            urls.forEach(url_ => {
+                const pathname = url.parse(url_).pathname;
+                const locale = pathname.split("/")[1];
+                allPossible[pathname] = locale;
```
wbamberg: I must be missing something, because I don't understand why `allPossible` isn't just an array of URIs.

scripts/linting/mdn-sitemap-compare.js
peterbe: I'm failing to find where we decided to use Node >=10. It's not in the `package.json` as an `engine`. 
Was it just something we chatted about and agreed in verbal only?

I'm actually OK with leaving this code here because it'll definitely work on Travis and the only kind of people who'd use and hack on this (similar to the scraper) are the "hard core" users like yourself. And we should make sure we have modern tools installed.

So I'd be tempted to just leave this comment be for now.

wbamberg: We didn't fully resolve it but it was discussed here: https://github.com/mdn/stumptown-content/issues/141. But the LTS is 12 now anyway, so using `matchAll` here is fine :).

scripts/linting/mdn-sitemap-compare.js
peterbe: I think I did it because I wanted to be able to do quick lookups. `!!someObj[key]` is much faster than `someArray.includes(key)`. 
But that's what `Set`s are for :)

peterbe:
@wbamberg I dropped the ball on this in that I still haven't addressed your comments. 

Now I've looked a bit deeper and addressed all your feedback. 
The reason `allPossible` is an object and not an array is so that you can do something like:
```
▶ node scripts/linting/mdn-sitemap-compare.js "fr"
/fr/docs/Web/HTML                                                                                  ✅
/fr/docs/Web/HTML/Element                                                                          ✅
/fr/docs/Web/HTML/Element/abbr                                                                     ✅
/fr/docs/Web/HTML/Element/video                                                                    🚫 (wrong case!)

In conclusion: 3 successful. 1 fails.

FYI, we have 4 docs and there are 6,957 possible (0.1%).
```

In this PR I also added this script to `.travis.yml` but it's failing. 
It's failing because of two reasons:

1. We have a typo on https://github.com/mdn/stumptown-content/blob/master/content/fr/html/reference/elements/video/video.md because the MDN URL is actually `Video` not `video`. 
2. There are a bunch of CSS reference docs like https://github.com/mdn/stumptown-content/blob/master/content/en-US/css/reference/properties/border/border.md that have a full absolute URL for the `mdn_url` value. I can fix that easily.

peterbe:
What are we going to do with the fact that it break TravisCI now? I wasn't really ready to merge this :)

wbamberg:
I'm making a PR to fix the CSS ones right now. The fr/video one should be fixed in the Wiki, which I am also trying to sort out.

I didn't know you weren't ready to merge it.

peterbe:
> I'm making a PR to fix the CSS ones right now. The fr/video one should be fixed in the Wiki, which I am also trying to sort out.
> 

I don't know what it means to "fixed in the Wiki" but I suspect it'll mean you'll need to fix it in stumptown-content too, no?

> I didn't know you weren't ready to merge it.

Sorry. I lost control on this. I still think this script it super useful but I didn't have a good solution to offer about how to tackle the fact that it broke things. I was hopeing to hear from you, depending on stuff you're more familiar with:

1. Can we fix all the pages that break?
2. Would you like that temporarily make this script NOT error but just print a warning in TravisCI?

I'm guessing you're going to gun for option 1.

wbamberg:
> > I'm making a PR to fix the CSS ones right now. The fr/video one should be fixed in the Wiki, which I am also trying to sort out.
> 
> I don't know what it means to "fixed in the Wiki" but I suspect it'll mean you'll need to fix it in stumptown-content too, no?

I'm not sure why. The URL is wrong in the Wiki but right in stumptown-content, I think.

Except! I tried moving:

https://wiki.developer.mozilla.org/fr/docs/Web/HTML/Element/Video

to 

https://wiki.developer.mozilla.org/fr/docs/Web/HTML/Element/video

and it complained, I thought because of a subpage at:

https://wiki.developer.mozilla.org/fr/docs/Web/HTML/Element/Video/playing

...which is just a redirect to https://wiki.developer.mozilla.org/fr/docs/Web/API/HTMLMediaElement/playing_event (it was created in the wrong place).

So I deleted https://wiki.developer.mozilla.org/fr/docs/Web/HTML/Element/Video/playing, and tried moving https://wiki.developer.mozilla.org/fr/docs/Web/HTML/Element/Video again. Now I get an error like:

Screen Shot 2019-12-13 at 11 13 10 AM

I assume there is supposed to be something listed after "suivants :". Maybe page move just doesn't work when you are only changing case? Oh man, I hate Kuma.

> > I didn't know you weren't ready to merge it.
> 
> Sorry. I lost control on this. I still think this script it super useful but I didn't have a good solution to offer about how to tackle the fact that it broke things. I was hopeing to hear from you, depending on stuff you're more familiar with:
> 
>     1. Can we fix all the pages that break?

Ha, see above. We can fix things in stumptown but I don't know about the Wiki.

>     2. Would you like that temporarily make this script NOT error but just print a warning in TravisCI?

The thing about warnings is, once you tolerate them you end up with hundreds of them, and then you never really know which is any are important.

peterbe:
> We can fix things in stumptown but I don't know about the Wiki.

You know better than I do about your status and intentions about scraping. 

Either way, if you just let it be busted in Kuma (MySQL) you'll still need to make a note somewhere in stumptown-content about this oddity. 

The ugly truth is that Kuma actually doesn't care:

Both `curl -I https://developer.mozilla.org/fr/docs/Web/HTML/Element/Video`  and `curl -I https://developer.mozilla.org/fr/docs/Web/HTML/Element/video` returns a 200 OK. But when it's written down in the https://developer.mozilla.org/sitemaps/fr/sitemap.xml as `Web/HTML/Element/Video`.

Ideally, think Kuma should correct the case. I.e. `curl https://developer.mozilla.org/en-US/docs/lEaRn/hTmL` should 301 redirect to `https://developer.mozilla.org/en-US/docs/Learn/HTML` otherwise you're basically suggesting that they are two different pages. 

(`<link rel="canonical" href="https://developer.mozilla.org/en-US/docs/Learn/HTML" >` is present when you go to https://developer.mozilla.org/en-US/docs/lEaRn/hTmL which tells the likes of Googlebot that that's the canonical spelling)

When we move to the jamstack, and upload a file called `en-US/docs/Web/Learn/HTML/index.html` what's going to happen to poor schmucks who try to use `/en-US/docs/lEaRn/hTmL`?
With S3 website hosting, I know it's going to trigger a 404. E.g. [`/en-US/docs/Web/HTML/Element/AuDiO/`](https://d27ve6vvlmvdm0.cloudfront.net/en-US/docs/Web/HTML/Element/AuDiO/)
I guess this is why Netlify keeps it forcibly simple and force everyone to always only use lowecase for all paths. 

I'll bring this up with Ryan and Gregor as a thing we simply need to be prepared for.

wbamberg:
If we care about case (which I would like us to), then `Web/HTML/Element/video` is right and `Web/HTML/Element/Video` is wrong. So ideally we would fix sitemap.xml to contain `Web/HTML/Element/video`. Is it possible to do this?

Otherwise maybe we should make this test case-insensitive?

peterbe:
> So ideally we would fix sitemap.xml to contain `Web/HTML/Element/video`. Is it possible to do this?

Not unless we change the slug in Kuma. I actually can't think of any reason NOT to do is. 
Can't we just edit it here: https://wiki.developer.mozilla.org/fr/docs/Web/HTML/Element/Video$edit
"Fragment d'URL en Français :" 
@SphinxKnight any objections? Or do you prefer to make it `Video` in `stumptown-content`?

This won't be the first time this happens. There'll be more cases where we "incorrectly" write down the `mdn_url` in stumptown-content. So I think the script is good as is for now.

wbamberg:
> Not unless we change the slug in Kuma. I actually can't think of any reason NOT to do is.
Can't we just edit it here: https://wiki.developer.mozilla.org/fr/docs/Web/HTML/Element/Video$edit
"Fragment d'URL en Français :"
@SphinxKnight any objections? Or do you prefer to make it Video in stumptown-content?

I tried doing this, and documented my failure here: https://github.com/mdn/stumptown-content/pull/156#issuecomment-565572480

peterbe:
I scraped so I had /fr/docs/Web/HTML/Element/video available on my local copy and it just worked. 
Screen Shot 2019-12-16 at 8 42 15 PM
No errors this time. 

I did it on https://wiki.developer.allizom.org/fr/docs/Web/HTML/Element/Video$edit too. 
It worked. Now, https://wiki.developer.allizom.org/fr/docs/Web/HTML/Element/video is the canonical URL. 
First I clicked "Modifier", then expand the "Traduire la description" and typed "video" where it said "Video". Then I pressed the "Publier et poursuivre l'édition" button.

SphinxKnight:
> @SphinxKnight any objections? Or do you prefer to make it `Video` in `stumptown-content`?
Well, no objection at all :) Thanks for @-ing and fixing this.

If I may be opinionated about titles and slugs for "guide" content, I think that reference pages should follow the case of the specs.

As time allows, I'll sweep existing page to check for their slug's case (in HTML / CSS / JS in French).

Issue #126, opened 2019-09-10
Support spec tables in HTML pages


wbamberg:
HTML reference pages in MDN (and most other reference pages) contain a table linking to relevant specification(s).

We didn't add this to the HTML element recipe when we first wrote them because it wasn't very clear which specifications we should include. But we need to have spec tables in Stumptown, so this issue is to work out how to specify spec tables in the recipes and add it to the data.

**Acceptance criteria:**
* all HTML recipes that need it have a spec table structure
* all HTML data that need it has spec table data

wbamberg:
For reference, here's the discussion doc on what spec tables should look like: https://docs.google.com/document/d/1eL8YtslVZAnIAGb7rcZGbvXndkcR7lb9rhDpvGpGaWs/edit?ts=5c763cb9#.

Elchi3:
Thanks for linking the discussion, Will! There are many ideas and issues in that document.

In BCD, we have spec_url along with mdn_url. This has landed for ECMAScript features, and Mike Smith has a fork that added more spec_urls here: https://github.com/w3c/browser-compat-data. That fork also added spec_urls to HTML features. I think I would like to try to upstream Mike's work into BCD, so that we can get spec_urls for HTML features through this route. Would you agree with that?
At a minimum, we can then take these URLs and simply reference the spec as the absolute minimum spec "table". I'd call this version 0. I think this would serve your acceptance criteria, do you think so, too? Also, "it wasn't very clear which specifications we should include", I agree, and I would double-check Mike's work as part of this. What he's done for ECMAScript features made a lot of sense (only link to the latest updated spec). So that question could be solved with this. As a plus point, all spec_urls become mass-editable.

In a follow-up, we could then experiment and get back to the ideas in the document you've linked and enrich the simple spec reference. For example, with spec status or other relevant bits for the future "spec table" (ideally we would user test these ideas). Some enrichments will need more data points than just the spec_url. Maybe something like a modern "specdata" that we currently have in KumaScript. I think that requires more work, but I think I would rather think about this stuff, when we know what the ideal spec table is and I belive only a dedicated project with user tests and having consulted some spec people would probably tell us. Like, it would probably require to create a 4-5 different spec tables from the ideas, test them with real users on the current MDN wiki pages for a few weeks and then decide. I wouldn't want get into this now, but please tell me if you think we should get into that.

wbamberg:
Thanks Florian!

[...]

PR #162, opened 2019-10-09
Add spec_urls to html docs

Elchi3:
This is the first part of https://github.com/mdn/stumptown-content/issues/126#issuecomment-539448613. It adds `spec_url` to the front matters in HTML docs.

The relevant, current, and non-historical spec that https://github.com/w3c/browser-compat-data uses, is https://html.spec.whatwg.org/multipage, so I've used that one here as well. There were only very few cases where I needed two specs (see iframe attributes, I think two specs make sense there).

While digging into this, I found a few things:

A) I think it's worth adding `spec_url` to attributes and not add all specs that define attributes to their element like done in https://github.com/w3c/browser-compat-data/blob/master/html/elements/iframe.json#L7. In this PR instead, the iframe `spec_url` is pointing to the iframe definition and then in the front matters of the attributes the respective attribute spec links are added. This avoids larger tables with comments like there are done on MDN now: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#Specifications. Even if we don't render out the `spec_url` of attributes, it feels like the much better place in the data.

B) Some attributes have been added by mistake. I was able to realize this when I was unable to find a `spec_url` for them. I've identified at least 3 cases:
1) Migration error: `open` attribute for the `dfn` element. No idea how it ended up there.
2) Mismatched properties of Web API vs. attributes of the HTML element. (e.g. `buffered` on video element seems to be actually an API property).
3) Outdated attributes that never had a real spec (e.g. `intrinsicsizes` on video). Here you could explicitly say "non-standard".

I wonder if we want `spec_url: non-standard`, because if you don't provide a `spec_url` it could either mean that the feature has no spec and is fact a non-standard thing, or it means that we weren't able to find the correct spec, or that we've made a mistake. Possibly by being explicit and stating `non-standard`, we could maybe make it harder to run into the error cases that I've described above as it makes you think about providing a `spec_url` like I was forced to in this PR. Just a thought, though.

This PR just adds the data without validation. I would do that next and see how much I got wrong, but let know if you prefer to have validation with this PR already.


wbamberg:
I think this looks great.

> I think it's worth adding spec_url to attributes

I can definitely see the value in this. It does make it impossible to build big spec tables showing spec status for this item and all its attributes all together (like the big BCD tables). If you wanted to render spec URL for attributes you'd then have to do it separately for each attribute, inside the documentation for that attribute. Which is maybe better anyway.

From a linting/architecture point of view, though, it's worth keeping in mind that attributes files are different from HTML element files or CSS property files because they don't map to MDN pages and don't have a recipe. They're "supplementary files" like code samples.

This means that (for example) for recipe-based docs (like HTML element files or CSS property files) we can write generic code that checks for the existence of `spec_url` based on the recipe. But we can't do the same for attributes files - checking the front matter in attributes files would be the responsibility of the special-purpose code that understands how to parse the `attributes` recipe property. And if we had some other supplementary files, for, say, individual CSS property keyword values, and they wanted `spec_url`, then checking that would be the responsibility of the special-purpose code that knew how to put together CSS property keyword values.

I don't know if this is a problem. Not for now, but might be a thing to keep in mind for the future.

> I wonder if we want `spec_url: non-standard`

I like the idea of being more explicit here.  If we do this then would `spec_url` be made mandatory? I'm worried that it will entail a lot of work tracking down missing specs if we do that. Perhaps we could support `non-standard` so we can represent that fact, but not make `spec_url` mandatory at least yet?

Elchi3:
> From a linting/architecture point of view, though, it's worth keeping in mind that attributes files are different from HTML element files or CSS property files because they don't map to MDN pages and don't have a recipe. They're "supplementary files" like code samples.

Okay, I see. I have the feeling that we might want to make these their own pages one day, but I don't think it would a big problem. For now, as you said as well, I'm not sure there is a problem. Probably not, so if you're fine with `spec_urls` in attributes then I am, too. :) From an authoring experience, it felt very right to me to add them to their front matters.

> I like the idea of being more explicit here. If we do this then would spec_url be made mandatory? I'm worried that it will entail a lot of work tracking down missing specs if we do that. Perhaps we could support non-standard so we can represent that fact, but not make spec_url mandatory at least yet?

I agree. I would also say lets add `non-standard` to the overall proposal now but still have `spec_url` as an optional thing. If you want, I can add `non-standard` to this PR, or add it in a follow-up. 
In any case, I will add it to the proposal in https://github.com/mdn/stumptown-content/issues/126#issuecomment-539448613 so that how we do specs in stumptown is documented somewhere for now (I should also move this whole comment to "stumptown docs" but I will wait for your "stumptown writer's guide").

peterbe:
How would you do linting on those? I can imagine some really basic linting where we check that it's a valid URL :)

wbamberg:
>  For now, as you said as well, I'm not sure there is a problem. 

Yeah, let's just keep an eye on it for now :).

> I can add non-standard to this PR, or add it in a follow-up.

Could you add it in this PR? Then it covers all the updates to content AFAICT.

Elchi3:
Okay, updated this.

#### 1) Non standard added to:

- html.elements.button.autocomple
- html.elements.canvas.moz-opaque
- html.elements.form.autocapitalize


#### 2) No specs added because I think these are actually global attributes and we should fix this:
`body` (see https://html.spec.whatwg.org/multipage/dom.html#global-attributes:event-handler-content-attributes)
 - `onblur`
 - `onerror`
 - `onfocus`
 - `onload`
 - `onredo`
 - `onresize`
 - `onundo`

`button` 
  - `autofocus` (see https://html.spec.whatwg.org/#the-autofocus-attribute:attr-fe-autofocus-7)


#### 3) I've deleted a few things:

`dfn`
  - `open` is not an attribute here (migration error)

`label`
 - `form` is a readonly API property of HTMLLabelElement. This is not a settable HTML attribute.

 `video`
 - `buffered` is a a readonly API property of HTMLMediaElement. This is not a settable HTML attribute.
 - `intrinsicsize` is non-standard but also the whole "explainer" spec is obsolete https://github.com/WICG/intrinsicsize-attribute

content/html/HTML.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: "HTML: Hypertext Markup Language"
 mdn_url: /en-US/docs/Web/HTML
+spec_url: https://html.spec.whatwg.org/multipage/
```
wbamberg: I hadn't thought of this, but having spec links on landing pages is really nice.

PR #167, opened 2019-10-11
Useful filenames

wbamberg:
This PR enables us to give docs files sensible names instead of having to call them "docs.md".

Specifically it removes the place this was hardcoded in build-json, which was the implementation of the `directory` method of building link lists.

In this method you provide a directory, like "content/html/reference/elements" and you get a list of all the pages directly under that directory. This used to work by just listing all subdirectories, then opening the file called "docs.md" in there.

Now it's a bit more involved, because it can't rely on the filename. It *does* expect that every subdirectory contains content for one and only one page. I think this seems like a good assumption. I wonder if we could also assume that the subdirectory can contain only one .md file. At the moment we can't, because of contributors.md, but I think we should have a different way to represent Wiki contributors.

I'm not massively in love with the code but it seems to work, and I've tested it using stumptown-renderer - the sidebars and landing pages in particular still seem fine.

Apart from the change in build-json, I also renamed all the pages under elements/reference.


peterbe:
I love it! "docs.md" never sat well with me. I couldn't get over the "plural sounding" of docS as if it's many documents. 

First of all, thank you for making every page a folder. That makes so much sense. 

A strong contender is to call it `index.md`. It doesn't actually have the same functionality like an index.html does in Nginx. But it has some pros and cons.

PR #169, opened 2019-10-14
Add initial linter

ddbeck:
OK, here is my much awaited linter for our content (for #90).

## What this thing does

If you invoke the linter with `npm run lint-md` (which replaces the earlier `remark` call), then this linter:

1. Collects and loads the recipes in `/recipes`.
1. Collects and identifies what I've termed "root docs" in the `/content` file tree. Root docs are those Markdown or `meta.yaml` files with a YAML value for `recipe`. (I figured that some frontmatter, prose, examples, and assets represents the smallest unit you might want to inspect, for symmetry with the files output from `build-json`.)
1. For each of those root docs:
   1. Parse the Markdown (or `meta.yaml` treated as the frontmatter for `prose.md`) into a remark (parsed Markdown) tree.
   1. Parse the YAML frontmatter (or its `meta.yaml` equivalent).
   1. From the frontmatter, get the recipe name, and, if valid, attach the recipe itself as some metadata on the tree.
   1. Run checks on the tree for deprecations and required sections.

## What this linter checks

In that process, the linter does a few checks and issues warnings or errors for:

* Using `meta.yaml` instead of Markdown frontmatter
* Setting recipe to a name not defined in `/recipes`
* Using a deprecated prose section (see #74 and #55)
* Failing to include a require prose section (see #74 and #55)

It doesn't check any of the other files yet. It doesn't know about examples or anything other than the regular prose.

## Implementation notes

I tried to do this in a very `unified` sort of way in the hopes that extending this linter will be non-maddening (c.f., the BCD linter). This gets us some pretty output and the option to generate JSON messages (for consuming into the GitHub Checks API, for example, or for editor integration).

This does require a bit of machinery to work, however, which makes for a large initial PR. Also, to demonstrate where you might plug in additional checks, I didn't put this together in the most performance-minded way. For example, the code in `deprecated-sections` and `missing-sections` walk the trees twice for the same subset of nodes; a more efficient and complex approach would run both these checks in the same traversal. For now though, I don't think there's a performance problem.

Finally, I've probably duplicated some code with the build scripts. I'm OK with this right now—I like the idea of playing with ways to process `/content` while we're still experimenting.


ddbeck:
One more thing: I don't currently set the exit status for the linter, because there are a number of warnings and (what I would expect to be) outright failures in the content. Look at the Travis output for details on what's not passing muster (and a pretty good example of the output).

scripts/linter/index.js
```diff
@@ -0,0 +1,92 @@
...
+            }
+        })
+        .use(missingSections);
+
+    const filePaths = await collectDocs();
+
+    const reportedFiles = [];
+
+    for (const fp of filePaths) {
+        if (fp.endsWith("meta.yaml")) {
```
wbamberg: I don't think we are expecting to have meta.yaml anywhere in the tree from now on. I see in this code that you are actually treating it as an error, but given that finding this error requires a bunch of special-casing in the linter, I think it would be better just to fix the three CSS files where it occurs:

```
content/css/reference/properties/border/meta.yaml
content/css/reference/properties/border-block-start-width/meta.yaml
content/css/reference/properties/transform/meta.yaml
```

...and remove the checking code from the linter.

scripts/linter/valid-recipes.js
```diff
@@ -0,0 +1,40 @@
...
+            if (hasRecipeInYaml(node)) {
+                const recipe = node.data.yaml.recipe;
+
+                if (validRecipes.includes(recipe)) {
+                    tree.data = tree.data || {};
+                    tree.data.recipe = recipes[recipe];
+                    tree.data.recipeName = recipe;
+                } else {
+                    const m = file.message(
+                        `"${recipe}" recipe was not found in /recipes`,
```
wbamberg: This errors for guide pages and landing pages, that have recipes named in front matter (`guide` and `landing-page`), but don't have YAML files, because they're not obviously amenable to the recipe.yaml model (because they don't have the same kind of structure as a reference page, which we can represent as a sequence of named sections).

Guide pages are just totally unstructured blobs of Markdown, with special bits (e.g. live samples) embedded in them at arbitrary places. Landing pages are a prose overview followed by a list of lists of links. So they're both treated as special cases, where the recipe name just tells the builder what to expect, without an actual recipe to consult.

I can think of about three options here:

* continue to live with this weirdness
* give guide pages and landing pages actual recipes, and use them. Now I think about this, we could perhaps write a useful recipe for landing pages. Still I think the guide page recipe would just be "some stuff". Another related problem is that at the moment, for reference pages, the recipe determines the sidebar (because all HTML element pages want the same sidebar), but guide pages and landing pages are generic across domains, so a "guide" recipe can't determine the sidebar.
* add some extra datum like "type", which could be ["guide"|"landing-page"|"reference"]. If type is "reference", expect "recipe" to exist.

FWIW I do expect that guide pages and landing pages would be the only pages that are not recipe-driven.

scripts/linter/missing-sections.js
```diff
@@ -0,0 +1,52 @@
...
+
+const ruleId = "stumptown-linter:missing-section";
+
+/**
+ * Re
+ */
+function requiredSections(recipe) {
+    const proseRequiredIngredients = recipe.body.filter(
+        ingredient =>
+            ingredient.startsWith && // ignore non-string ingredients
```
wbamberg: Why not `typeof(ingredient) === 'string'`?

scripts/linter/valid-recipes.js
ddbeck: Can we have recipe files that don't specify a `body`? Or even essentially emtpy recipe files that say, `# read whatever.js to see how this recipe is handled`?

It would be nice to have a single place to see what the valid recipes are, even if some recipes are special cases that only have code that describes how they work.

I had a thought during this PR (and my attempts to migrate some HTML content) that it'd be nice to specify, for example, expected frontmatter (whether required or optional). The recipe seems like a natural place for that, so I can imagine uses for recipe files even for kinds of recipes that don't get built the same way as ordinary reference pages.

ddbeck:
Will, I just pushed a bunch of changes based on your comments. An overview of what's new and different:

* `meta.yaml` files have moved into corresponding Markdown files as frontmatter.
* The code for finding `meta.yaml` and their special handling is now gone. There's just one linting routine now.
* Unparsable frontmatter now produces prettier error messages. Removing `meta.yaml` files caused me to discover how frontmatter can break and the linter handles this better now.
* The linter now consumes _all_ Markdown files indiscriminately, though we don't yet check anything for Markdown files that don't have a `recipe` set. This was prompted by your suggestion to not use regexes for finding recipes—it turned out to be easier to just ignore stuff when we don't know the correct recipe than it is to check the files in advance.
* That said, I did add a check to flag files that _should_ have a recipe (thanks to the changes in #168).
* I removed `heading-utils` and deduplicated the code that used those utils. See `scripts/linter/plugins/slugify-sections.js` for details.
* I moved the unified plugins into `scripts/linter/plugins`.
* I fixed a few small things you pointed out (e.g., `typeof ingredient` or unnecessary `await`s) and several I discovered on my own (an incomplete comment, for example).

I think the only thing I haven't made any attempt to cope with is the question about recipes that don't have a corresponding file in `recipes` (https://github.com/mdn/stumptown-content/pull/169#discussion_r335212182).

scripts/linter/valid-recipes.js
wbamberg: > Can we have recipe files that don't specify a body? Or even essentially emtpy recipe files that say, # read whatever.js to see how this recipe is handled?

It would be nice to have a single place to see what the valid recipes are, even if some recipes are special cases that only have code that describes how they work.

Yes, I think this is probably a better solution than a secret exception like we have at the moment. Part of the recipe thing, for me, is meta-documentation.

As I said, we could also perhaps make landing pages properly recipe-driven.

But this isn't a linter issue (just an issue that the linter uncovered :) so let's keep the linter behaviour as it is, to remind us to come back to it.

scripts/linter/missing-sections.js
wbamberg: Yes, I've been there. I guess my feeling is that if `typeof(ingredient) === 'string'` doesn't work properly, that's JavaScript's problem not mine :).

scripts/linter/collect-docs.js
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+}
+
+/**
+ * Checks if there's a recipe name in `filePath`.
+ */
+async function isRootDoc(filePath) {
+    const file = await fsPromises.readFile(filePath);
+    return /^recipe: [\w-]+$/gm.test(file);
```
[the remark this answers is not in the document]
wbamberg: By definition, the correct answer (at the moment) to "is this file the place to start building a thing" is "this is a Markdown file containing front matter, that contains a `recipe` property".

Because that's how build-json decides what it's going to try to build. So ideally the linter should use exactly the same definition.

But I don't think this is a blocker for the linter at this point anyway.

>  The naive approach read each Markdown file, walked its tree, and parsed its YAML two times: once to identify the Markdown file as having a recipe and once to run checks on it. Avoiding the repetition would've introduced a lot of complexity that the regex sidestepped.

Yes, build-json is super-inefficient and eventually that will become a problem. I wonder if vfiles could help us here?

scripts/linter/collect-docs.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Walks the content directory tree (`start`) yielding Markdown files.
+ */
+async function* walkContent(start = "content") {
+    const files = await fsPromises.readdir(start, { withFileTypes: true });
+
+    for (const f of files) {
+        if (f.name.endsWith(".md")) {
+            yield path.join(start, f.name);
+        } else if (f.isDirectory()) {
```
peterbe: The order is wrong. If there's a folder called `content/stuff/foo.md/` it'd yield that instead of digging deeper into it.

scripts/linter/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+    const reportedFiles = [];
+
+    for (const fp of filePaths) {
+        const file = await vfile.read(fp);
+        await markdownParser().process(file);
+        reportedFiles.push(file);
+    }
+
+    console.error(report(reportedFiles, { quiet: true }));
```
peterbe:
```
▶ npm run lint-md

> stump@0.1.0 lint-md /Users/peterbe/dev/MOZILLA/MDN/stumptown-content
> node scripts/linter

content/css/reference/properties/border/border.md
            warning  "css-property" recipes requires section "example-syntax"     missing-section           stumptown-linter
            warning  "css-property" recipes requires section "see-also"           missing-section           stumptown-linter
            warning  "css-property" recipes requires section "short-description"  missing-section           stumptown-linter
            warning  "css-property" recipes requires section "syntax-overview"    missing-section           stumptown-linter

content/css/reference/properties/border-block-start-width/border-block-start-width.md
            warning  "css-property" recipes requires section "example-syntax"     missing-section           stumptown-linter
            warning  "css-property" recipes requires section "see-also"           missing-section           stumptown-linter
            warning  "css-property" recipes requires section "short-description"  missing-section           stumptown-linter
            warning  "css-property" recipes requires section "syntax-overview"    missing-section           stumptown-linter

content/css/reference/properties/transform/transform.md
            warning  "css-property" recipes requires section "example-syntax"     missing-section           stumptown-linter
            warning  "css-property" recipes requires section "see-also"           missing-section           stumptown-linter
            warning  "css-property" recipes requires section "short-description"  missing-section           stumptown-linter
            warning  "css-property" recipes requires section "syntax-overview"    missing-section           stumptown-linter

content/html/HTML.md
  1:1-14:4  error    "landing-page" recipe was not found in /recipes              frontmatter-valid-recipe  stumptown-linter

content/html/guides/Applying_color/Applying_color.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/html/reference/elements/base/base.md
            warning  "html-element" recipes requires section "see_also"           missing-section           stumptown-linter

content/html/reference/elements/datalist/datalist.md
            warning  "html-element" recipes requires section "overview"           missing-section           stumptown-linter

content/html/reference/elements/dd/dd.md
            warning  "html-element" recipes requires section "overview"           missing-section           stumptown-linter

content/html/reference/elements/elements.md
   1:1-9:4  error    "landing-page" recipe was not found in /recipes              frontmatter-valid-recipe  stumptown-linter

content/html/reference/elements/figcaption/figcaption.md
            warning  "html-element" recipes requires section "overview"           missing-section           stumptown-linter

content/html/reference/elements/html/html.md
            warning  "html-element" recipes requires section "overview"           missing-section           stumptown-linter

content/html/reference/elements/iframe/iframe.md
            warning  "html-element" recipes requires section "see_also"           missing-section           stumptown-linter

content/html/reference/elements/legend/legend.md
            warning  "html-element" recipes requires section "overview"           missing-section           stumptown-linter

content/html/reference/elements/table/table.md
            warning  "html-element" recipes requires section "overview"           missing-section           stumptown-linter

content/learn/html/Introduction_to_HTML/Advanced_text_formatting/Advanced_text_formatting.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/Creating_hyperlinks/Creating_hyperlinks.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/Debugging_HTML/Debugging_HTML.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/Document_and_website_structure/Document_and_website_structure.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/Getting_started/Getting_started.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/HTML_text_fundamentals/HTML_text_fundamentals.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/Introduction_to_HTML.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/Marking_up_a_letter/Marking_up_a_letter.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/Structuring_a_page_of_content/Structuring_a_page_of_content.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

content/learn/html/Introduction_to_HTML/The_head_metadata_in_HTML/The_head_metadata_in_HTML.md
   1:1-6:4  error    "guide" recipe was not found in /recipes                     frontmatter-valid-recipe  stumptown-linter

33 messages (✖ 13 errors, ⚠ 20 warnings)

▶ echo $?
0
```

So even though there were 13 errors it still finished OK. 
I'm not really sure how `vfile-reporter` works (or even what it is!) but perhaps we leave that and on the line below this one, we filter `reportedFiles` for things that are actual errors and do something like `process.exitCode = reportedFiles.filter(v => v.error).length;`

peterbe:
Perhaps best done as a follow-up feature but can the linter please also catch files that lack a last newline character. 
See https://github.com/mdn/stumptown-content/pull/184/files for example. Lots of files that don't end with a `\n`. A fair amount of people have their editors set to always add one, so if they edit one of these files, once checked in, they'll produce a diff that's outside of where they actually made a change.

peterbe:
It appears this did NOT catch https://github.com/mdn/stumptown-content/issues/191

Should it? 
I.e. does this linter pluck apart the front-matter and checks if the paths that it mentions there are resolved correctly?

wbamberg:
> It appears this did NOT catch #191
> 
> Should it?
> I.e. does this linter pluck apart the front-matter and checks if the paths that it mentions there are resolved correctly?

Not at present, no. See [the above comment](https://github.com/mdn/stumptown-content/pull/169#pullrequestreview-302243002): "At some point we'll want the linter to dig down into individual ingredients, like attributes files and live samples". This PR is the first version of the linter, it doesn't try to catch errors at that level of detail. But we will.

peterbe:
So we will use the linter for this ...eventually?
I started [a fix](https://github.com/mdn/stumptown-content/compare/mdn:master...peterbe:191-open-files-case-sensitively) which works but if the `lint-md` script is a better place to put it we can just do that instead. ...once this lands.

wbamberg:
> So we will use the linter for this ...eventually?

Yes, I think so.

scripts/linter/collect-docs.js
ddbeck: Good catch. Fixed.

scripts/linter/collect-docs.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Return an array of paths to Markdown files.
+ */
+async function collectDocs() {
+    const candidateDocs = [];
+
+    for await (const candidateDoc of walkContent()) {
+        candidateDocs.push(candidateDoc);
+    }
```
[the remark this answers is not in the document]
ddbeck: Interesting results for globbing. Since, as you say, it doesn't matter, I won't rewrite it to be synchronous, but I will keep that in mind for the future.

> It's not wrong but I don't understand this JS. The `walkContent` is asynchronous, right? So what was the point of then turning it into a big array?

`collectDocs` used to do some other things beside squashing `walkContent`, but when we dropped `meta.yaml` files, building an array was all that was left. I've unrolled that bit. Now, `walkDocs` yields each Markdown file asynchronously (had to rename the module for clarity and consistency though).

scripts/linter/index.js
ddbeck: [`vfile`](https://github.com/vfile/vfile) is `unified`'s library for doing stuff with files. It's pretty cool. I'm using it here for attaching (and pretty printing) error messages, though it does other things (for example, tracks file renames). It also provides an API for getting stats about those reports, which I've used to set the exit code. Thank you!

ddbeck:
@peterbe I'd prefer to just merge it, if that's OK with you and @wbamberg. It'd be a useful thing immediately, for migrating HTML elements and such. If we don't want the failing CI, we could comment out the `process.exit()` line and I could follow up with another PR to turn it back on, when the problems are fixed.

peterbe:
It's your call. As long as you have a plan to fix them :)
Truth be told, we already have other [more pressing problems in Travis](https://github.com/mdn/stumptown-content/issues/176).

peterbe:
@ddbeck Feel free to merge.

PR #173, opened 2019-10-17
include query used for bcd resolution

peterbe:
Basically, this reveals the BCD query (e.g. `html.elements.link.rel`) back in the packaged JSON so that the renderer can extract that primary name (e.g. `rel`) when it builds up the rows.


peterbe:
Renderer companion: https://github.com/mdn/stumptown-renderer/pull/186

PR #177, opened 2019-10-18
Add spec data and build it to packaged output

Elchi3:
This is the next part for https://github.com/mdn/stumptown-content/issues/126.

It does these things: 
- adds a `data/specifications.json` file containing allowed specs with their titles.
- exposes a new `specifications` property to the packaged JSON of pages and HTML attributes.

I wonder if we want `data/specifications.yaml` instead. It would allow us to comment on our spec choices.

### DIFF
--- a/scripts/build-json/build-specs.js
+++ b/scripts/build-json/build-specs.js
@@ -1,6 +1,6 @@
 const fs = require('fs');
 const path = require('path');
-
+const yaml = require('js-yaml');
 const { ROOT } = require('./constants');
 
 function packageSpecs(specUrls) {
@@ -11,8 +11,7 @@ function packageSpecs(specUrls) {
   }
 
   const dataDir = path.resolve(ROOT, 'content', 'data');
-  const file = fs.readFileSync(path.join(dataDir, 'specifications.json'), 'utf8');
-  const specMap = JSON.parse(file);
+  const specMap = yaml.safeLoad(fs.readFileSync(path.join(dataDir, 'specifications.yaml'), 'utf8'));
 
   function findTitle(specUrl) {
     let title = ''
--- a/scripts/build-json/build-specs.js
+++ b/scripts/build-json/build-specs.js
@@ -21,7 +21,7 @@ function packageSpecs(specifications) {
       }
     });
     if (title === '') {
-      throw new Error(`Domain for "${specification}" not found in data/specifications.json`);
+      throw new Error(`Domain for "${specification}" not found in data/specifications.yaml`);
     }
     return title;
   }
--- a/scripts/build-json/build-specs.js
+++ b/scripts/build-json/build-specs.js
@@ -1,0 +1,52 @@
+const fs = require('fs');
+const path = require('path');
+
+const { ROOT } = require('./constants');
+
+function buildSpecs(specUrls) {
+
+  // spec_url is optional, return nothing so there is no specifications property
+  if (!specUrls) {
+    return;
+  }
+  // spec_url can be just "non-standard", route it through in that case
+  if (specUrls === 'non-standard') {
+    return specUrls;
+  }
+
+  const dataDir = path.resolve(ROOT, 'content', 'data');
+  const file = fs.readFileSync(dataDir + '/specifications.json', 'utf8');
+  const specMap = JSON.parse(file);
+
+  function findTitle(specUrl) {
+    let title = ''
+    Object.entries(specMap).forEach(([key, value]) => {
+      if (specUrl.includes(key)) {
+        title = value;
+      }
+    });
+    if (title === '') {
+      console.error(`Domain for "${specUrl}" not found in data/specifications.json`);
+    }
+    return title;
+  }
+
+  // For authoring convenience, you can provide one or more specs
+  if (!Array.isArray(specUrls)) {
+    specUrls = [specUrls];
+  }
+
+  let specifications = [];
+  specUrls.forEach(specUrl => {
+    specifications.push({
+      url: specUrl,
+      title: findTitle(specUrl)
+    });
+  });
+
+  return specifications;
+}
+
+module.exports = {
+  buildSpecs
+}
--- a/scripts/build-json/build-specs.js
+++ b/scripts/build-json/build-specs.js
@@ -11,7 +11,7 @@ function packageSpecs(specUrls) {
   }
 
   const dataDir = path.resolve(ROOT, 'content', 'data');
-  const file = fs.readFileSync(dataDir + '/specifications.json', 'utf8');
+  const file = fs.readFileSync(path.join(dataDir, 'specifications.json'), 'utf8');
   const specMap = JSON.parse(file);
 
   function findTitle(specUrl) {
@@ -22,7 +22,7 @@ function packageSpecs(specUrls) {
       }
     });
     if (title === '') {
-      console.error(`Domain for "${specUrl}" not found in data/specifications.json`);
+      throw new Error(`Domain for "${specUrl}" not found in data/specifications.json`);
     }
     return title;
   }
--- a/scripts/build-json/build-specs.js
+++ b/scripts/build-json/build-specs.js
@@ -3,43 +3,43 @@ const path = require('path');
 const yaml = require('js-yaml');
 const { ROOT } = require('./constants');
 
-function packageSpecs(specUrls) {
+function packageSpecs(specifications) {
 
-  // spec_url can be just "non-standard", route it through in that case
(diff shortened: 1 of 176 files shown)


wbamberg:
Thanks for the PR, @Elchi3 .

I would have expected the `spec_url` item to be listed in the recipe for HTML elements and HTML input elements as an optional `meta.` ingredient. This would help indicate to authors that it is expected in the front matter. It would then be processed in [`processMetaIngredient`](https://github.com/mdn/stumptown-content/blob/master/scripts/build-json/build-recipe-page-json.js#L9), and would end up in the JSON's `body`, which is what the renderer looks in [when it puts the document together](https://github.com/mdn/stumptown-renderer/blob/master/client/src/document.js#L151). Things outside `body` are more for metadata, although I could see this might be a tricky distinction to make sometimes.

We might want to keep the same name in the source and in the built JSON (i.e. rather than transform `spec_url` into `specifications`). This isn't needed but is something we have conventionally done, and might make it easier to understand how a part of a rendered document maps back to a thing in a recipe and a source MD file. If we wanted to do that, then `specifications` would perhaps be a better name throughout than `spec_url`?

I think perhaps we should use YAML for specifications.json, just because we use that for other files in stumptown. It seems inconsistent to use both languages.

scripts/build-json/build-specs.js
```diff
@@ -0,0 +1,52 @@
...
+  if (!specUrls) {
+    return;
+  }
+  // spec_url can be just "non-standard", route it through in that case
+  if (specUrls === 'non-standard') {
+    return specUrls;
+  }
+
+  const dataDir = path.resolve(ROOT, 'content', 'data');
+  const file = fs.readFileSync(dataDir + '/specifications.json', 'utf8');
```
wbamberg: I think this should use `path.join()`.

scripts/build-json/build-specs.js
```diff
@@ -0,0 +1,52 @@
...
+
+  function findTitle(specUrl) {
+    let title = ''
+    Object.entries(specMap).forEach(([key, value]) => {
+      if (specUrl.includes(key)) {
+        title = value;
+      }
+    });
+    if (title === '') {
+      console.error(`Domain for "${specUrl}" not found in data/specifications.json`);
```
wbamberg: I think this should throw an exception.

Elchi3:
> I would have expected the `spec_url` item to be listed in the recipe for HTML elements and HTML input elements as an optional `meta.` ingredient. This would help indicate to authors that it is expected in the front matter. It would then be processed in [`processMetaIngredient`](https://github.com/mdn/stumptown-content/blob/master/scripts/build-json/build-recipe-page-json.js#L9), and would end up in the JSON's `body`, which is what the renderer looks in [when it puts the document together](https://github.com/mdn/stumptown-renderer/blob/master/client/src/document.js#L151). Things outside `body` are more for metadata, although I could see this might be a tricky distinction to make sometimes.

Thanks for the pointers, Will! My stumptown picture gets more and more complete whenever I get your feedback! :)
I've added a commit that attempts to do it this way. I think this has the downside that it won't package up the specs for HTML element attributes, for example. It seems like we're only exposing to consumers what we're showing in the rendered docs then. Also, in the export, `specifications` isn't available as a main property, like `abbr.specifications`, but you'll have to get it from the `body` tree: `abbr.body["type=specifications"].value`. Maybe I'm too focused on how we expose things, though. This works and if it is consistent with the rest, then I guess I'm happy with it. 
Also, thanks for the pointer to the renderer code, I will have to dive into that soon, but from briefly looking at it, yes, I can see how this will then get picked up.

Elchi3:
> We might want to keep the same name in the source and in the built JSON (i.e. rather than transform `spec_url` into `specifications`). This isn't needed but is something we have conventionally done, and might make it easier to understand how a part of a rendered document maps back to a thing in a recipe and a source MD file. If we wanted to do that, then `specifications` would perhaps be a better name throughout than `spec_url`?

Yes, I'd be in favor of being consistent. I think `spec_url `didn't make sense to me anymore when I made it an object with two properties `title` and `url`. Do you think it make sense in this PR to change all the `spec_url` to `specifications` in all the `.md` files?

Elchi3:
> I think perhaps we should use YAML for specifications.json, just because we use that for other files in stumptown. It seems inconsistent to use both languages.

Done! It seems like an improvement to me, too. JSON isn't very friendly to edit and we have the infrastructure to use yaml here easily.

wbamberg:
> Yes, I'd be in favor of being consistent. I think `spec_url `didn't make sense to me anymore when I made it an object with two properties `title` and `url`. Do you think it make sense in this PR to change all the `spec_url` to `specifications` in all the `.md` files?

Yes. Sorry!

wbamberg:
> > I would have expected the `spec_url` item to be listed in the recipe for HTML elements and HTML input elements as an optional `meta.` ingredient. This would help indicate to authors that it is expected in the front matter. It would then be processed in [`processMetaIngredient`](https://github.com/mdn/stumptown-content/blob/master/scripts/build-json/build-recipe-page-json.js#L9), and would end up in the JSON's `body`, which is what the renderer looks in [when it puts the document together](https://github.com/mdn/stumptown-renderer/blob/master/client/src/document.js#L151). Things outside `body` are more for metadata, although I could see this might be a tricky distinction to make sometimes.
> 
> Thanks for the pointers, Will! My stumptown picture gets more and more complete whenever I get your feedback! :)
> I've added a commit that attempts to do it this way. I think this has the downside that it won't package up the specs for HTML element attributes, for example. It seems like we're only exposing to consumers what we're showing in the rendered docs then. 

Is there any reason we can't add `specifications` to the attributes JSON , inside compose-attributes.js? I think it is fine for the packaged JSON to include content that the renderer has chosen not to show. (Now I look at compose-attributes, I see we are not exposing BCD in there either, but the same applies I think).

> Also, in the export, `specifications` isn't available as a main property, like `abbr.specifications`, but you'll have to get it from the `body` tree: `abbr.body["type=specifications"].value`. Maybe I'm too focused on how we expose things, though. This works and if it is consistent with the rest, then I guess I'm happy with it.

Definitely since you last were closely involved with this, it has got more aligned with the renderer and less abstract. I don't necessarily think this is a bad thing but the risks are that it could be too limiting for other consumers.

I'd be very happy to get your perspective on these questions in London!

Elchi3:
> > Yes, I'd be in favor of being consistent. I think `spec_url `didn't make sense to me anymore when I made it an object with two properties `title` and `url`. Do you think it make sense in this PR to change all the `spec_url` to `specifications` in all the `.md` files?
> 
> Yes. Sorry!

Added a commit that does the rename. I also updated our spec for specs in stumptown :-) https://github.com/mdn/stumptown-content/issues/126#issuecomment-539448613

Elchi3:
> Is there any reason we can't add specifications to the attributes JSON , inside compose-attributes.js? I think it is fine for the packaged JSON to include content that the renderer has chosen not to show. (Now I look at compose-attributes, I see we are not exposing BCD in there either, but the same applies I think).

I think there isn't a reason we can't add it, but I removed it since I wasn't sure what our plan is for exposing things. I thought it maybe is "only expose what the renderer will render". For BCD, we're again doing what is best for the renderer: Instead of putting the BCD in piece meals on each attribute, we have a JSON all together in one spot to render out a table that has the attribute compat data in several rows.

> Definitely since you last were closely involved with this, it has got more aligned with the renderer and less abstract. I don't necessarily think this is a bad thing but the risks are that it could be too limiting for other consumers.

Yes, being less abstract is okay, I think. When I was first imaging this, I was more thinking about exposing a structure that describes things we're documenting, so e.g. HTML elements have a spec or compat data and so it is accessible like `abbr.specifications` and `abbr.attributes.foo.compat_data`. 
Now it seems more like this structure is modeled after rendered pages. So, in a way, the structure doesn't describe the HTML abbr element directly, but it is framed as an MDN page where you access your meat in `abbr.body["type=specifications"].value` and `abbr.body["type=browser_compatibility"].value.data.foo.__compat`. This is fine given we need to build for mdn the web site, it is just something that I had to realize. Lacking additional content consumers, I don't know if this less abstract structure is limiting at all. It might not be (like the good ol' saying: "Pine doesn't care _where_ it is, as long as it is _somewhere_!")

> I'd be very happy to get your perspective on these questions in London!

Yes, happy to continue with this in London. I don't want to derail PRs with these thoughts.

peterbe:
@Elchi3 I filed https://github.com/mdn/stumptown-renderer/issues/191 for the follow-up to the renderer.

PR #181, opened 2019-10-22
upgrade jsdom

peterbe:


package.json
```diff
@@ -26,7 +26,7 @@
     "dependencies": {
         "gray-matter": "4.0.2",
         "js-yaml": "3.13.1",
-        "jsdom": "^15.1.1",
+        "jsdom": "15.2.0",
```
peterbe: I don't like ranges but this time I got the range notation removed because I used `npm install --save jsdom@15.2.0`.

wbamberg:
I don't understand what the reason for this change is, can you explain?

peterbe:
> I don't understand what the reason for this change is, can you explain?

The upgrade; no reason for the newer version but I noticed that it was using a range rather than an exact version. That makes sure every dev gets the same version which might also be *part* of the reason why our package-lock.json changes simply from doing `npm install`.

Issue #171, opened 2019-10-17
Additional prose sections are not getting titles rendered


wbamberg:
Reference pages can contain predefined prose sections as defined in the recipe (like "Overview" or "Accessibility concerns"), and also additional sections with custom names. For example, the `iframe` page wants a section called "Scripting".

When stumptown-renderer renders these sections is does not render the "Scripting" H2. This is because it only renders headings for prose sections that contain an `id` property: https://github.com/mdn/stumptown-renderer/blob/master/client/src/document.js#L156, and stumptown-content doesn't set `id` for custom sections.

I'm not sure whether this is an error in content or renderer.

Content doesn't set the `id` because it thinks the purpose of `id` is  is for consumers to extract a particular named section. For example, for an editor to extract all the `short_description` sections from the HTML elements. That makes it useful only for predefined sections, not for sections with random names.

But I think renderer sees `id` as setting the [`id`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) attribute, and it wants to set that attribute for H2 headings. One issue with this understanding is that if `id` is needed for H2 headings, why isn't it also needed for all the other heading levels? Or is it in fact needed, and this should also be present in the stumptown-content HTML?

PR #184, opened 2019-10-23
Added HTML guides

wbamberg:
This PR adds support for all the HTML guide pages: that is, those listed in https://github.com/mdn/stumptown-content/issues/159. I also updates "Guides.yaml" so they all show up in the sidebar.

Except for two:
* https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats: has moved and morphed into a kind of Web API landing page
* https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache: is obsolete

It's not perfect but maybe it's good enough for our purposes. In general I'm a bit uncertain how for to go in tweaking the source.

Some things that came up:

## Heading IDs

Pages like https://developer.mozilla.org/en-US/docs/Web/HTML/Date_and_time_formats#Examples include internal links, that don't work because the markup doesn't have the requisite IDs. I'm not sure what to do about that - whether adding them is the renderer's job or content's. But I think we should have them. Note that this is related to https://github.com/mdn/stumptown-content/issues/171.

## Specifications

[As Florian said](https://github.com/mdn/stumptown-content/pull/178#pullrequestreview-305277317), we might want to support `specifications` in guide pages. There are 3 pages here that have a "Specifications" sections:

* https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types#Specifications
* https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#Specifications
* https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes#Specifications

For now I left the first as a raw table and removed the others, kind of randomly. It would be easy to add support to guide pages of course, but every bit of extra stuff adds complexity, and I'm not sure that this very formal presentation is really appropriate for guide pages. It feels like these pages are trying to be reference pages, and we should consider refactoring the content. But that's not in scope for this PR IMO.

## `dl` elements and multicolumn

These pages use [`<dl>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#Conceptual_differences) and [multicolumn layout](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#Elements) in places. These seem to be getting converted mostly as just `<ul>`. We could extend the Markdown processing to support `dl` if we think it is important enough. Multicolumn might be harder. If we supported [`link_list`](https://github.com/mdn/stumptown-content/blob/master/scripts/build-json/build-link-lists.js) as a directive (which we should IMO) we could render them multicol, and that might be a good option.

## spell-md

I have a love-hate relationship with the spellcheck linter. Probably 70% hate, 30% love. It is a real grind updating .spelling and it seems to need a lot of updates, and there are lots of things that are annoying. Like apparently you have to add "domain.org" for every domain. I don't know if there better ways to deal with this kind of situation. On the other hand it does find genuine errors, and that's great. I wonder if we should not run it in CI, but just have it as a discretionary tool. Or maybe as Peter said, make it generate warnings only.


peterbe:
> Or maybe as Peter said, make it generate warnings only.

After all, that's what Google docs, Word, and some fancy textareas do. You know in your heart that `textareas` is fine in this context so you just ignore it. 

Screen Shot 2019-10-22 at 10 04 23 PM

What might be truly interesting is investing in other more advanced machine learning based services such as the ones Grammarly is based on. 
Then, after you submit a pull request it might post some comments on the PR about a poor grammar. 

Screen Shot 2019-10-22 at 10 06 39 PM

Elchi3:
> It's not perfect but maybe it's good enough for our purposes. In general I'm a bit uncertain how for to go in tweaking the source.

Yes, I think there is a tendency for us to always also wanting to improve the original content. It's hard to judge sometimes.

> ## Heading IDs
> 
> Pages like https://developer.mozilla.org/en-US/docs/Web/HTML/Date_and_time_formats#Examples include internal links, that don't work because the markup doesn't have the requisite IDs. I'm not sure what to do about that - whether adding them is the renderer's job or content's. But I think we should have them. Note that this is related to #171.

I agree, I'm a big fan of section IDs on headings to be able to link there. Doesn't need to block this PR, but I would like to see it happen.

>  ## Specifications
>
> [As Florian said](https://github.com/mdn/stumptown-content/pull/178#pullrequestreview-305277317), we might want to support `specifications` in guide pages.
> [...]
> I'm not sure that this very formal presentation is really appropriate for guide pages

Having thought about this some more, I think that the specification section is not complex enough (yet) and so if you want to mention specs somewhere in a guide, just do it manually without any formalism. On that topic: You've included an old style spec table with spec status and comments. I'm not sure if it is still the best way to communicate what we want to communicate there given we've now decided to get rid of spec status and spec comments. I could imagine it still to be a table, but to concentrate more on the aspect of what is specified where and leave out the now obsoleted formalism. Does that make sense? Could change in this PR or follow up on it.

>  It feels like these pages are trying to be reference pages, and we should consider refactoring the content. But that's not in scope for this PR IMO.

Yes, I already had the impression for the first two guide pages you've migrated. Strong hints that HTML attributes could be own pages (which I think I would be in favor of given we now had multiple signs that it would make sense). However, agree it is not necessary to refactor in this PR.

> ## `dl` elements and multicolumn
> 
> These pages use [`<dl>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#Conceptual_differences) and [multicolumn layout](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#Elements) in places. These seem to be getting converted mostly as just `<ul>`. We could extend the Markdown processing to support `dl` if we think it is important enough. Multicolumn might be harder. If we supported [`link_list`](https://github.com/mdn/stumptown-content/blob/master/scripts/build-json/build-link-lists.js) as a directive (which we should IMO) we could render them multicol, and that might be a good option.

I think this is fine for now. It seems to me like we want directives (or just css classes?) for this kind of multicol or other defined layout. If we want to avoid having layouting mixed with content, I guess it could be separated in build-json, so that consumers can get content without layout annotations (but maybe it isn't a big deal anyway, especially in unstructured guide pages). Not blocking this PR.

> ## spell-md
>
> I have a love-hate relationship with the spellcheck linter

:-) Open to improve it or to make it warning only, but I'm also curious to see the implementation from the Unified system to see if it is any better then.

wbamberg:
> You've included an old style spec table with spec status and comments. I'm not sure if it is still the best way to communicate what we want to communicate there given we've now decided to get rid of spec status and spec comments. I could imagine it still to be a table, but to concentrate more on the aspect of what is specified where and leave out the now obsoleted formalism. Does that make sense? Could change in this PR or follow up on it.

Yes, I do agree, but feel that this is also "content refactoring" that in general tried to avoid here, as I wouldn't know where to stop. I feel more inclined to do content refactoring for reference pages, partly because I just have to, to make them consistently fit structures.

PR #187, opened 2019-10-23
Use prettier for all scripts

### LINKED ISSUE #186: Use prettier for all scripts
Let's just copy the implementation from stumptown-renderer.

peterbe:
Fixes #186

I simply copied all the ideas over from `stumptown-content`. 

Note, I want to wait with this until https://github.com/mdn/stumptown-content/pull/169 lands. And I expect I need to rebase this because of the `package-lock.json` changes. 

Also, this won't pass CI until I go ahead and re-format all the existing files. It's a good start.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
         "prettier": "1.18.2",
         "pretty-quick": "2.0.0"
     },
+    "prettier": {},
     "husky": {
         "hooks": {
             "pre-commit": "pretty-quick --staged"
--- a/package.json
+++ b/package.json
@@ -13,10 +13,10 @@
         "url": "https://github.com/mdn/stumptown-content/issues"
     },
     "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -59,9 +59,13 @@
     },
     "devDependencies": {
         "eslint": "^6.5.1",
-        "prettier": "1.18.2"
+        "husky": "3.0.9",
+        "prettier": "1.18.2",
+        "pretty-quick": "2.0.0"
     },
-    "prettier": {
```
ddbeck: Can we still have a `prettier` key in `package.json` or an empty `.prettierrc` file in the repo? I use the presence of [prettier config](https://prettier.io/docs/en/configuration.html) to turn on automatic formatting in my editor.

peterbe: Done.

peterbe:
I'll start processing individual files that I don't think are affected by the linter.

PR #197, opened 2019-10-26
Writers guide

wbamberg:
Here's a first draft "writers guide".

It's supposed to help orient writers in stumptown, and also go into the details of how the various structures in stumptown are defined.

It should help you understand the authoring interface, and also writing it down should help expose some of the problems and inconsistencies it has, and places we can clean it up. I've added notes like **Note: a thing** in places where it seems to me we want to change things.

I'm interested in knowing if you think this doc makes sense as an explanation of the "authoring interface", and I also I'm definitely interested in knowing about things in here that surprise or concern you or are just too complicated.

### DIFF
--- a/project-docs/stumptown-writers-guide.md
+++ b/project-docs/stumptown-writers-guide.md
@@ -32,11 +32,11 @@ Almost all of this document focuses on the stumptown-content repository, since t
 
 ## The /content directory
 
-All content created by writers lives in the ./content directory. This directory contains a collection of Markdown and YAML files. Markdown is used for prose content, and YAML is used to tell stumptown where it can find other content (such as BCD) and how it should combine prose content with other content. The Markdown files often also contain YAML in the form of front matter.
+All content created by writers lives in the /content directory. This directory contains a collection of Markdown and YAML files. Markdown is used for prose content, and YAML is used to tell stumptown where it can find other content (such as BCD) and how it should combine prose content with other content. The Markdown files often also contain YAML in the form of front matter.
 
-Not all Markdown files under ./content correspond directly to MDN pages, but most of them do. Those that do all contain YAML front matter which contains a property named `recipe`. This property tells stumptown what kind of page this content wants to be.
+Not all Markdown files under /content correspond directly to MDN pages, but most of them do. Those that do all contain YAML front matter which contains a property named `recipe`. This property tells stumptown what kind of page this content wants to be.
 
-Let's look at one of these Markdown files. It's located at ./content/html/reference/elements/b/docs.md, and is the content for the [HTML `<b>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b) page:
+Let's look at one of these Markdown files. It's located at /content/html/reference/elements/b/docs.md, and is the content for the [HTML `<b>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b) page:
 
 ```
 ---
--- a/project-docs/stumptown-writers-guide.md
+++ b/project-docs/stumptown-writers-guide.md
@@ -1,5 +1,19 @@
 # Writers' guide to stumptown
 
+**Table of contents:**
+* [stumptown-content and stumptown-renderer](#stumptown-content-and-stumptown-renderer)
+* [The content directory](#the-content-directory)
+* [Recipe files](#recipe-files)
+* [Meta ingredient details](#meta-ingredient-details)
+* [Landing pages](#landing-pages)
+* [Guide pages](#guide-pages)
+* [Related content](#related-content)
+* [Chapter lists](#chapter-lists)
+* [Specifications](#specifications-1)
+* [Contributors](#contributors)
+
+## stumptown-content and stumptown-renderer
+
 This document describes stumptown from the point of view of a technical writer.
 
 We can think of the MDN documentation landscape like this:
--- a/project-docs/stumptown-writers-guide.md
+++ b/project-docs/stumptown-writers-guide.md
@@ -402,7 +402,7 @@ If `height` is omitted, then stumptown treats this example as just a static exam
 
 **Note: we have also considered having a less implicit way to signal whether or not an example is live (https://github.com/mdn/stumptown-content/issues/45 ).**
 
-The rest of this file is a prose description of the example.
+The rest of the "description.md" file is a prose description of the example.
 
 #### example.\*
 
@@ -681,7 +681,7 @@ Items listed in the `specifications` front matter property must match one of the
 specifications: https://html.spec.whatwg.org/multipage/semantics.html#the-abbr-element
 ```
 
-**Note: I don't see a reason why this adds a scheme, but the scheme is not present in "specifications.yaml".**
+**Note: I don't see a reason why the front matter property includes a scheme, but the scheme is not present in "specifications.yaml".**
 
 ## Contributors
 
--- a/project-docs/stumptown-writers-guide.md
+++ b/project-docs/stumptown-writers-guide.md
@@ -396,7 +396,7 @@ height: 192
 ---
 ```
 
-If `height` is included, then stumptown treats this example as a live sample, and `height` is the height of the "Result" iframe in pixels (the iframe always gets full page width).
+If `height` is included, then stumptown treats this example as a live sample, and `height` is the height of the "Result" iframe in pixels (the iframe always gets full page width). The renderer will render the example like a live sample in the current MDN (that is, displaying the sources, and the output in an iframe).
 
 If `height` is omitted, then stumptown treats this example as just a static example.
 
@@ -687,13 +687,6 @@ specifications: https://html.spec.whatwg.org/multipage/semantics.html#the-abbr-e
 
 ## Contributors
 
-Markdown files in stumptown that can be built into pages must have a "contributors.md" file in the same directory. The content of this file is a comma-separated list of Markdown links to MDN user profiles, one for each MDN use who contributed to this page:
+Markdown files in stumptown that can be built into pages must have a "contributors.md" file in the same directory. This lists contributions made to the original page in the Kuma Wiki. This file is created when the page is first migrated from the Wiki and should then be treated as read only.
 
-```
-[mfuji09](/en-US/profiles/mfuji09),
-[jswisher](/en-US/profiles/jswisher),
-[vadimhp](/en-US/profiles/vadimhp),
-[SphinxKnight](/en-US/profiles/SphinxKnight)
-```
-
-**Note: this is the most hacky and least acceptable piece of stumptown. In particular, I think that legacy contributor information should not live in /content, because it is read-only: authors should never need to be aware of it. But this is only one of its problems.**
+**Note: this approach to attributing legacy contributions is very likely to change in future, and contributors.md will be moved out of /content.**
--- a/project-docs/stumptown-writers-guide.md
+++ b/project-docs/stumptown-writers-guide.md
@@ -175,7 +175,7 @@ short_title: <b>
 mdn_url: /en-US/docs/Web/HTML/Element/b
 ```
 
-**Note: It's arguable whether these three items should actually be in the recipe, but they are applicable to every page, so it seemed redundant.**
+**Note: It's arguable whether these three items should actually be in the recipe, but they are applicable to every page, so it seemed redundant. Filed as https://github.com/mdn/stumptown-content/issues/203.**
 
 See [meta ingredient details](#meta-ingredient-details) for detailed descriptions of each meta ingredient.
 
@@ -287,7 +287,7 @@ In this example, there are two element-specific attributes, found at "../input/a
 
 Paths may be given as an absolute path from the stumptown root directory, or a path relative to the file containing the ingredient.
 
-**Note: I think we should remove `global`, since everyone always has it and it always has the same value.**
+**Note: I think we should remove `global`, since everyone always has it and it always has the same value. Filed as https://github.com/mdn/stumptown-content/issues/202.**
 
 Each element-specific attribute is described in its own Markdown file, with its own syntax.
 
@@ -331,6 +331,8 @@ The H1 heading must be in `code format` and its content must be the name of the 
 # `crossorigin`
 ```
 
+**Note: We should not use code format here or in the "Values" bit. Filed as https://github.com/mdn/stumptown-content/issues/198.**
+
 Prose immediately after this describes the attribute.
 
 Some attributes, such as the `crossorigin` attribute above, can take special enumerated values. For such attributes, the description should be followed by an H2 with the text "Values":
@@ -357,7 +359,7 @@ After this there is an string describing the type of this attribute. The followi
 
 An object used to specify static or live samples to include in the page.
 
-**Note: this does not specify interactive examples, because we'd have to rewrite the live sample system if we wanted to do that.**
+**Note: this does not specify interactive examples, because we'd have to rewrite the interactive example system if we wanted to do that.**
 
 It is given as an array of paths to directories. Each directory contains a single example:
 
@@ -368,7 +370,7 @@ examples:
     - examples/multiple-source-elements
 ```
 
-**Note: it looks like this is using a different path convention from attributes. We should be consistent here.**
+**Note: it looks like this is using a different path convention from attributes. We should be consistent here. Filed as https://github.com/mdn/stumptown-content/issues/204.**
 
 Examples will be included in the rendered output in the order given here: so for example you may want to present the basic example first in the array, then more complex ones.
 
@@ -379,7 +381,7 @@ Each example is specified as a directory containing four files, one mandatory an
 -   "example.css"
 -   "example.js"
 
-**Note: we don't yet support "hidden" sources One suggestion was just to call them "hidden.css" etc.**
+**Note: we don't yet support "hidden" sources One suggestion was just to call them "hidden.css" etc. Filed as https://github.com/mdn/stumptown-content/issues/205.**
 
 #### description.md
 
--- a/project-docs/stumptown-writers-guide.md
+++ b/project-docs/stumptown-writers-guide.md
@@ -118,7 +118,7 @@ body:
 
 This describes what must be present (and in some cases what _may_ be present) in all HTML element pages, and the order in which items must appear. Each distinct type of page in MDN - CSS property pages, Web API method pages, JS function pages - will have one of these files, although some of them will be very similar.
 
-The first item, `related_content`, points to a special YAML file that describes other pages related to this one. It's used to build a sidebar for this page (but could have other uses too).
+The first item, `related_content`, points to a special YAML file that describes other pages related to this one. It's used to build a sidebar for this page (but could have other uses too). See [Related content](#related-content).
 
 The other item, `body`, describes the structure of the page content. Each property under here names a piece of the page.
 
@@ -177,7 +177,7 @@ mdn_url: /en-US/docs/Web/HTML/Element/b
 
 **Note: It's arguable whether these three items should actually be in the recipe, but they are applicable to every page, so it seemed redundant.**
 
-See [meta ingredient details](#meta-ingredients-details) for detailed descriptions of each meta ingredient.
+See [meta ingredient details](#meta-ingredient-details) for detailed descriptions of each meta ingredient.
 
 ### Prose ingredients
 
@@ -220,7 +220,7 @@ This was formerly known as the
 
 Heading names are mapped to identifiers by being lowercased and having spaces converted to underscores. So we can see here that the `<br>` docs include all the three mandatory prose sections, and none of the optional sections.
 
-As for `prose.*`, this is a mechanism for authors to include custom sections with any heading they choose. For example, the `<iframe>` page wants a heading "Scripting". The `<br>` page, though, doesn't make use of this.
+As for `prose.*`, this is a mechanism for authors to include custom sections with any heading they choose. For example, the `<iframe>` page wants a heading "Scripting".
 
 ## Meta ingredient details
 
--- a/project-docs/stumptown-writers-guide.md
+++ b/project-docs/stumptown-writers-guide.md
@@ -30,7 +30,7 @@ The "stumptown-renderer" repository implements the most important consumer of st
 
 Almost all of this document focuses on the stumptown-content repository, since this is where authors spend their time. Authors do not need to understand what's going on inside stumptown-renderer, and they do not need to understand the JSON structures that stumptown-content builds. However, they will need to use stumptown-renderer so they can see what their content looks like when rendered as MDN pages.
 
-## The ./content directory
+## The /content directory
 
 All content created by writers lives in the ./content directory. This directory contains a collection of Markdown and YAML files. Markdown is used for prose content, and YAML is used to tell stumptown where it can find other content (such as BCD) and how it should combine prose content with other content. The Markdown files often also contain YAML in the form of front matter.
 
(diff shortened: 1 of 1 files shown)


project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+        - ../input/attributes/autocomplete.md
+        - ../input/attributes/name.md
+    global: /content/html/global_attributes
+```
+
+In this example, there are two element-specific attributes, found at "../input/attributes/autocomplete.md" and "../input/attributes/name.md", relative to the directory containing this front matter.
+
+Paths may be given as an absolute path from the stumptown root directory, or a path relative to the file containing the ingredient.
+
+**Note: I think we should remove `global`, since everyone always has it and it always has the same value.**
```
Elchi3: I agree!

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+Note that the front matter also contain three items not in the recipe:
+
+```yaml
+title: "<b>: The Bring Attention To element"
+short_title: <b>
+mdn_url: /en-US/docs/Web/HTML/Element/b
+```
+
+**Note: It's arguable whether these three items should actually be in the recipe, but they are applicable to every page, so it seemed redundant.**
```
Elchi3: I think we don't need these 3 in the recipes. Maybe we should see them as things you'll always need no matter what kind of content you're creating, so I think I would see them de-coupled from individual page recipes, but as "basic requirements" for any page.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+It is given as an array of paths to directories. Each directory contains a single example:
+
+```yaml
+examples:
+    - examples/simple-example
+    - examples/audio-element-with-video-element
+    - examples/multiple-source-elements
+```
+
+**Note: it looks like this is using a different path convention from attributes. We should be consistent here.**
```
Elchi3: I like this one. (without any "../" or "./")

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+Examples will be included in the rendered output in the order given here: so for example you may want to present the basic example first in the array, then more complex ones.
+
+Each example is specified as a directory containing four files, one mandatory and three optional:
+
+-   "description.md" - the only mandatory one
+-   "example.html"
+-   "example.css"
+-   "example.js"
+
+**Note: we don't yet support "hidden" sources One suggestion was just to call them "hidden.css" etc.**
```
Elchi3: I like hidden.{css|js|html}!

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+Paths may be given as an absolute path from the stumptown root directory, or a path relative to the file containing the ingredient.
+
+**Note: I think we should remove `global`, since everyone always has it and it always has the same value.**
+
+Each element-specific attribute is described in its own Markdown file, with its own syntax.
+
+#### Element-specific HTML attribute syntax
+
+Here's an example HTML attribute file:
```
Elchi3: Personally, I don't like that much how attribute md files are different from say the main element md file, but this is where we're at right now. Can you maybe describe a bit how these md files are different and not really "pages". It would maybe also help me to better understand why we made it this way.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+From MDN's own meta-docs:
+
+> A landing page serves as a menu, of sorts, for its subpages, and is therefore primarily a navigation page. A landing page layout is typically used for the root page of a tree of pages about a particular topic. It opens with a brief summary of the topic, then presents a structured list of links to its subpages, and optionally, additional material that be useful to the reader... The list of subpages can be generated automatically...
+
+So typically they are made of some prose content on some high-level topic like HTML, followed by a collection of links to the pages comprising the docs for this topic.
+
+Stumptown supports landing pages using Markdown files with YAML front matter. Although their front matter does list a `recipe: landing-page` property, the recipe itself isn't used in building the page.
+
+**Note: we should just make landing pages recipe-driven like the reference pages.**
```
Elchi3: I think this would be useful so that landing pages appear consistent throughout MDN.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+* "embed-compat": this tells stumptown that the guide wants to embed a BCD table at this point. Its argument is a BCD query string like "html.elements.audio".
+
+We will probably add support for more directives in future, not too many. One likely candidate would be "embed-link-list".
+
+Directives can be inserted anywhere in guide pages. They can only be used in guide pages.
+
+Directives look a lot like KumaScript macros, because they are essentially the same.
+
+**Note: we have an open issue to use a less terrible syntax for directives: https://github.com/mdn/stumptown-content/issues/103 .**
```
Elchi3: The "Liquid" syntax that (I think) originated from Jekyll, is the one I have seen the most in md files so far. https://jekyllrb.com/docs/liquid/

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+wicg.github.io/priority-hints: 'Priority Hints'
+```
+
+Items listed in the `specifications` front matter property must match one of the URLs in this file. They may add extra path elements and fragments, to point to individual items within a specification:
+
+```
+specifications: https://html.spec.whatwg.org/multipage/semantics.html#the-abbr-element
+```
+
+**Note: I don't see a reason why the front matter property includes a scheme, but the scheme is not present in "specifications.yaml".**
```
Elchi3: We can add it. I have no strong opinion about it. :)

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+Markdown files in stumptown that can be built into pages must have a "contributors.md" file in the same directory. The content of this file is a comma-separated list of Markdown links to MDN user profiles, one for each MDN use who contributed to this page:
+
+```
+[mfuji09](/en-US/profiles/mfuji09),
+[jswisher](/en-US/profiles/jswisher),
+[vadimhp](/en-US/profiles/vadimhp),
+[SphinxKnight](/en-US/profiles/SphinxKnight)
+```
+
+**Note: this is the most hacky and least acceptable piece of stumptown. In particular, I think that legacy contributor information should not live in /content, because it is read-only: authors should never need to be aware of it. But this is only one of its problems.**
```
Elchi3: I agree! Maybe it should scrape into /legacy/* with a bunch of other read-only things we will need to have to preserve history?

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+## The ./content directory
+
+All content created by writers lives in the ./content directory. This directory contains a collection of Markdown and YAML files. Markdown is used for prose content, and YAML is used to tell stumptown where it can find other content (such as BCD) and how it should combine prose content with other content. The Markdown files often also contain YAML in the form of front matter.
+
+Not all Markdown files under ./content correspond directly to MDN pages, but most of them do. Those that do all contain YAML front matter which contains a property named `recipe`. This property tells stumptown what kind of page this content wants to be.
+
+Let's look at one of these Markdown files. It's located at ./content/html/reference/elements/br/docs.md, and is the content for the [HTML `<b>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b) page:
```
Elchi3:
```suggestion
Let's look at one of these Markdown files. It's located at ./content/html/reference/elements/b/docs.md, and is the content for the [HTML `<b>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b) page:
```

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+#### example.\*
+
+The example.html, example.css, and example.js files contain the example sources themselves.
+
+### specifications
+
+This property lists specifications for the item.
+
+This is given as one of:
```
a2sheppy: Two points:

* What about IETF specs in RFCs? We don't currently list these in our spec database, but do sometimes link to them in spec tables. We should decide how to handle these and then write that down in here.

* We should have some written rule about the boundary between what's got an entry in the spec table and what doesn't. We do sometimes document things which are "specified" in non-official documents such as community drafts. For instance, there has been discussion about providing some documentation for things like EPUB (https://www.w3.org/publishing/epub32/epub-spec.html), whose spec is a community group publication rather than an "official" W3C spec.

project-docs/stumptown-writers-guide.md
wbamberg: "less terrible syntax" was really a reference to the joke "{{{{{" syntax (which is deliberately silly, to remind us to reconsider the current syntax). I don't actually feel very strongly about the particular syntax here.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+## Title of another list of Links
+
+* [link text 3](link target 3): short description for the link 3
+* [link text 4](link target 4): short description for the link 4
+...
+```
+
+**Note: we don't currently support short descriptions for guide pages. But we should.**
+
+Each "link list" object may be specified in one of three ways:
```
a2sheppy: Need a fourth way: external link, defined by a title and a url.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+...
+
+## Title of another list of Links
+
+* [link text 3](link target 3): short description for the link 3
+* [link text 4](link target 4): short description for the link 4
+...
+```
+
+**Note: we don't currently support short descriptions for guide pages. But we should.**
```
a2sheppy: For the single page types of link list objects, we should allow the addition of an optional "description" property to override the description that would otherwise be fetched from the target page.

project-docs/stumptown-writers-guide.md
a2sheppy: Definitely; will also make it easier to build the site when these are recipe driven.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+The Markdown part is of course the guide's prose. To support the "special content" that guides want to embed, stumptown has the concept of "directives". A directive looks like:
+
+```
+{{{{{embed-example:examples/specifying-colors-in-stylesheets}}}}}
+```
+
+We currently support two directives:
+
+* "embed-example": this tells stumptown that the guide wants to embed a live sample at this point. Its argument is "examples/specifying-colors-in-stylesheets", which is a path to a directory containing an example. The content of this directory is just the same as the directories references by the `examples` front matter property, with a mandatory "description.md" and optional "example.[html|css|js]" files.
```
a2sheppy: How are these presented? As interactive examples? Or are they presented more like old-style "live" examples? Or some new way?

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+* "embed-compat": this tells stumptown that the guide wants to embed a BCD table at this point. Its argument is a BCD query string like "html.elements.audio".
+
+We will probably add support for more directives in future, not too many. One likely candidate would be "embed-link-list".
```
a2sheppy: +1 for `embed-link-list`.

Do we need an `embed-media` directive to embed a video or audio object in the page? Possibly also `embed-youtube` or such. I ask because there's nothing in this document about how to embed video.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+Directives can be inserted anywhere in guide pages. They can only be used in guide pages.
```
a2sheppy: Why only guide pages? I ask because I strongly suspect we will come to regret that limitation eventually. Especially if it winds up being the mechanism used to embed media files and the like.

project-docs/stumptown-writers-guide.md
a2sheppy: Looks a *lot* like how we used to embed DekiScript back on MindTouch.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+Items listed in the `specifications` front matter property must match one of the URLs in this file. They may add extra path elements and fragments, to point to individual items within a specification:
+
+```
+specifications: https://html.spec.whatwg.org/multipage/semantics.html#the-abbr-element
+```
+
+**Note: I don't see a reason why the front matter property includes a scheme, but the scheme is not present in "specifications.yaml".**
+
+## Contributors
```
ddbeck: Maybe we could not talk about this at all? Or at least say, "There's historic contributor data in `contributors.md`. Please don't modify these files."

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+    * Elements
+        * ...all HTML element pages
+    * Global attributes
+        * ...all HTML global attribute pages
+
+These related content files are referenced using the `related_content` property. This may appear either in recipes (for page types that all want the same related content collection), or in the front matter section of individual pages (for page types, like guides and landing pages, that don't all want the same related content collection).
+
+**Note: we will need to think about how to extend this grammar for more complex sidebars.**
+
+**Note: we should probably abandon this "infinite levels of sections" thing here and perhaps restrict ourselves to three levels (by not allowing children to contain children).**
```
a2sheppy: Agreed.

project-docs/stumptown-writers-guide.md
wbamberg: This part of stumptown hasn't changed since Portland. Back then we designed the format of attributes files as an attempt to balance between something that could express a structure ("Type", "Values" etc) with something that would be reasonably nice to write and to read.

One of the things about attributes files is that you have a lot of structure (e.g. "Values"), with bits of pure prose attached to bits of the structure (e.g. a description of an individual value).

There are general alternative approaches, like:
* express the structure as a YAML object, and embed the prose descriptions as attributes of the object. But I think we don't really want to embed blobs of Markdown inside YAML objects (or maybe we do?). It tends to push writing towards form-filling.
* express the structure as a YAML object, and point to the prose descriptions as separate Markdown files. But this will give us very fragmentary docs.
* ?

I'm very happy to consider other options, but not sure what they are. I also think this will get interesting when we start to deal with other types of pages that have similar features (e.g. functions, which have a very clear structure (params, returns, exceptions) with bits of pure prose attached.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+## Chapter lists
+
+A "chapter list" is just a way to provide a reusable list of links to pages, presented in a particular order. For example, a sidebar might want to list all the HTML guide pages in a curated order, with the most basic first, and more advanced concepts later. So the sidebar specification could provide its own list of pages to include. But an HTML landing page might also want to include the same list, in the same order. This is what chapter lists enable you to do.
+
+A chapter list is a YAML file with the following contents:
```
a2sheppy: It would be nice if we could have the ability to divide the chapter list up into titled sections, to be rendered like

**The basics**
- Introducing HTML
- Adding styles
- Showing pictures

**Forms and inputs**
* Adding forms to pages
* Input type guide
... etc...

wbamberg:
@Elchi3 , @ddbeck , @a2sheppy  - thanks for all the comments! I should have said this before, but: I think it would help my sanity if you can distinguish *comments on the doc* from *comments on stumptown*. I want the doc to describe what stumptown is currently, not what it might aspire to be.

Comments of the first sort should potentially prevent me from merging this PR.

Comments of the second sort should most likely result in us filing separate issues, maybe after a bit of discussion on this PR. For those we should also perhaps add these little "**Note:**" bits here, just to remind readers that this is expected to change.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+After this there is an string describing the type of this attribute. The following are valid values for this string:
+
+"String", "Number", "Boolean", "URL", "RegExp".
+
+### examples
+
+An object used to specify static or live samples to include in the page.
+
+**Note: this does not specify interactive examples, because we'd have to rewrite the live sample system if we wanted to do that.**
```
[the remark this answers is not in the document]
wbamberg: Sorry, yes, that is a typo. Should be " because we'd have to rewrite the **interactive example system** if we wanted to do that."

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+```
+
+**Note: it looks like this is using a different path convention from attributes. We should be consistent here.**
+
+Examples will be included in the rendered output in the order given here: so for example you may want to present the basic example first in the array, then more complex ones.
+
+Each example is specified as a directory containing four files, one mandatory and three optional:
+
+-   "description.md" - the only mandatory one
+-   "example.html"
```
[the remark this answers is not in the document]
wbamberg: At the moment I don't intend to support multiple chunks of code with prose interleaving.

I don't think it is worth the complexity. If we do find it's needed, we will look into adding it. In general though I really want stumptown to be a much simpler system than Kuma/KumaScript, and really to justify every bit of complexity we add to it.

>  I find it very useful to be able to break up the JavaScript in particular and explain it in smaller pieces.

I'd suggest we use code comments for this.

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+
+The file may contain some YAML front matter, with the following two properties, both optional:
+
+-   `browser_compatibility`: syntax is the same as the `browser_compatibility` ingredient
+-   `specifications`: syntax is the same as the `specifications` ingredient.
+
+The H1 heading must be in `code format` and its content must be the name of the attribute:
+
+```md
+# `crossorigin`
```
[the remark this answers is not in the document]
wbamberg: Filed as https://github.com/mdn/stumptown-content/issues/198

project-docs/stumptown-writers-guide.md
```diff
@@ -0,0 +1,697 @@
...
+### examples
+
+An object used to specify static or live samples to include in the page.
+
+**Note: this does not specify interactive examples, because we'd have to rewrite the live sample system if we wanted to do that.**
+
+It is given as an array of paths to directories. Each directory contains a single example:
+
+```yaml
+examples:
```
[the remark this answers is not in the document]
wbamberg: Well, that's a choice for the renderer :), and this doc intentionally doesn't talk about how stumptown-content gets rendered. But actually, yes, in stumptown-renderer they end up looking like live samples, with all the sources listed, then the result in an iframe.

But we could imagine other options too: for example I think we could render them as embedded CodePens if we wanted.

project-docs/stumptown-writers-guide.md
wbamberg: Filed as https://github.com/mdn/stumptown-content/issues/202.

project-docs/stumptown-writers-guide.md
wbamberg: Firstly because it makes everything more complex. Secondly, because it's part of the concept of reference pages that they can be quite formally structured, and letting people put random things like this wherever they like undermines this.

Mostly though, because it really breaks the idea that the content can be easily machine-accessed. At the moment in stumptown, a section of prose is just a section of relatively simple HTML. So it's possible, for example, for a consumer to fetch just the "Accessibility concerns" section of the doc, and display it in the context of an accessibility inspector. But what if this section contains an "embed-compat" directive? Well:
* the JSON output could include the directive, like our ?raw endpoint. Then the consumer has to understand how to process directives.
* the JSON output could include the rendered BCD table. Then stumptown-content is determining how the content ought to be rendered, which we really want to avoid.
* the JSON output could break up "Accessibility concerns" into the prose pieces and the interleaved BCD. This is what we do for guide pages. But then there isn't an "Accessibility concerns" at all.
* ?

> I ask because I strongly suspect we will come to regret that limitation eventually.

One advantage stumptown has over Kuma is that Kuma had to anticipate what the likely use cases were going to be. That's really hard and means you necessarily end up building lots of things you don't really need (but then find it hard to remove for various reasons). But stumptown is dealing with a really mature content base. So we know, pretty much, what the use cases are. It doesn't seem very likely that people will start writing radically new types of JS docs at this point.

project-docs/stumptown-writers-guide.md
wbamberg: Filed as https://github.com/mdn/stumptown-content/issues/203.

project-docs/stumptown-writers-guide.md
wbamberg: Filed as https://github.com/mdn/stumptown-content/issues/204.

project-docs/stumptown-writers-guide.md
wbamberg: Filed as https://github.com/mdn/stumptown-content/issues/205.

project-docs/stumptown-writers-guide.md
wbamberg: That's up to the renderer. stumptown-renderer presents them as live samples. But it could just as easily make them CodePen embeds, say.
Not interactive examples (CSS interactive examples have quite a different structure).

project-docs/stumptown-writers-guide.md
wbamberg: How many pages use this? In MDN we have 76 pages (en-US) that use EmbedYoutube. They are almost all in devtools, which won't be in stumptown. So I don't see a current need for this.

project-docs/stumptown-writers-guide.md
wbamberg: Why doesn't the second option "an explicit list of pages" cover this? I think it would be OK for these to be external pages.

project-docs/stumptown-writers-guide.md
wbamberg: I don't think we should do this, without a much better reason. I don't want to start embedding markup in YAML, and I think it would get hard to maintain.

project-docs/stumptown-writers-guide.md
wbamberg: Filed https://github.com/mdn/stumptown-content/pull/206 for this.

wbamberg:
Thanks very much @Elchi3 , @ddbeck , and @a2sheppy for such thoughtful comments! I think I haven't got to all of them yet but will try to tomorrow.

> I think I would like to see this split up a bit in the future, so that the writer's guide could talk about stumptown-content infrastructure generally and then about page types and page building blocks more specifically in separate writer's guide md files.

Yes, definitely. Especially the rather more formal specs about the different ingredients should be split out. But for now I also wanted to have a one-stop place for people to start understanding it all, without having to jump around too much.

project-docs/stumptown-writers-guide.md
wbamberg: I've filed https://github.com/mdn/stumptown-content/issues/215 for this.

project-docs/stumptown-writers-guide.md
wbamberg: @a2sheppy , please file these as separate issues.

project-docs/stumptown-writers-guide.md
wbamberg: I think we should leave this until and unless we have a definite need for it.

project-docs/stumptown-writers-guide.md
wbamberg: I've updated this section.

wbamberg:
I'm merging, since it's docs-only and has at least one r+. I think it will do more good in the repo than in a PR. If there are outstanding issues let's file follow-ups.

PR #206, opened 2019-10-29
Use a recipe to build landing pages

wbamberg:
This PR changes build-json so it uses a recipe to build landing pages. That leaves guide pages as the only page types that are not recipe-driven.

I wanted to remove any references to particular page types, so in the end I had to I made some changes to the css-property recipe, and eventually patched the code to ignore "css-property" pages, because the content for these page types isn't fixed up yet.

This does break stumptown-renderer :( because the previous code built landing pages as an array of "link_list" things, while this code builds them as a single "link_lists" thing that contains an array of the lists.

It would be possible to make this change without breaking stumptown-renderer, but I think the approach chosen here is simpler and more expressive for authors (I played around with things like `link_list*` which could mean zero or more `link_list` things, but it seemed like more syntax for authors to have to understand).

The change in stumptown-renderer is simple. I tested it with a change like:

```
$ git diff
diff --git a/client/src/document.js b/client/src/document.js
index fcd15b4..112367e 100644
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -202,9 +202,13 @@ function RenderDocumentBody({ doc }) {
       // https://github.com/mdn/stumptown-content/issues/106
       console.warn("Don't know how to deal with info_box!");
       return null;
-    } else if (section.type === "link_list") {
+    } else if (section.type === "link_lists") {
       return (
-        <LinkList title={section.value.title} links={section.value.content} />
+        <>
+        {section.value.map((link_list, i) => (
+          <LinkList key={i} title={link_list.title} links={link_list.content} />
+        ))}
+        </>
       );
     } else {
       console.warn(section);
```
...and it seemed to work fine.


wbamberg:
I don't understand why the build is failing here :(.

ddbeck:
Is it possible you're on a Mac, Will? Here's my theory:

`/content/html/guides/Block-level_elements/Block-level_elements.md` contains `{{{{{embed-example:examples/Block-level_example}}}}}`. Note the last bit: `examples/Block-level_example`, with a lower-case `_example`. The actual path to the file is `examples/Block-level_Example`, with an upper-case `_Example`. By default, the macOS file system is case insensitive so it works on my machine and—I'm guessing—yours. Meanwhile, Travis CI runs Linux, where the file system is case sensitive and the build fails.

If my guess is right, pushing a change from `{{{{{embed-example:examples/Block-level_example}}}}}` to `{{{{{embed-example:examples/Block-level_Example}}}}}` will fix that particular failure. And if I'm right, maybe we should consider using only lower-case file names.

wbamberg:
That worked, thank you @ddbeck . And yes, I agree about lower-case names. This is using random case because the scraper preserves the case when it dismembers live samples, but we ought to lower-case them.

scripts/build-json/build-page-json.js
```diff
@@ -29,6 +28,22 @@ function writeToFile(json, elementPath) {
...
+  } else {
+    const recipePath = path.join(__dirname, '..', '..', 'recipes', `${recipeName}.yaml`);
+    if (!fs.existsSync(recipePath)) {
+      throw new Error(`Not a supported recipe: ${recipeName}`);
+    }
+    recipe = yaml.safeLoad(fs.readFileSync(recipePath, 'utf8'));
+    recipeCache[recipeName] = recipe;
+  }
+  return recipe;
+}
```
peterbe: Perhaps a nit but I think you should use `in` like this instead:
```javascript
if (recipe in cache === false) {
    const recipePath = path.join(__dirname, '..', '..', 'recipes', `${recipeName}.yaml`);
    if (!fs.existsSync(recipePath)) {
      throw new Error(`Not a supported recipe: ${recipeName}`);
    }
    recipeCache[recipeName] = yaml.safeLoad(fs.readFileSync(recipePath, 'utf8'));
}
return cache[recipe];
```

wbamberg:
> I don't agree with your suggested (impending) diff on the renderer. I think it should be:

That was just illustrative, to check that a small change like that would work.

> I think you should use in like this instead

This was in the original code (I just made it into a separate function), but I've made the change anyway, it is nicer.

peterbe:
I'm more than happy to jump onto this in the renderer as soon as you merge it.

peterbe:
Also, this PR edits a lot of build-json .js scripts. Once this lands I'll make a PR to prettier those files.

wbamberg:
> I'm more than happy to jump onto this in the renderer as soon as you merge it.

Thanks Peter!

Issue #77, opened 2019-07-25
"Content sourcemaps"


peterbe:
The user story isn't crystal clear yet but something I think would be super engaging is to be able to refer to the source when you're reading something. 

Suppose you're reading the "Usage notes" on the `<video>` page and you spot a typo, or you found a bug in a code example, then it would be great to know exactly where the content came from. Some people might want just the file path of the .md file and some might just prefer to go straight to the HTML URL in GitHub. 

I'm referring to what Readthedocs does with it's "Edit in GitHub" link for example. 

It's an opportunity to engage new contributors but it also needs to be powerful enough for core contributors.

peterbe:
At the moment, technically the renderer has everything it needs. The `cli` knows the path of the file it picks up. But it's the .json file. So if the filepath is `stumptown/packaged/html/reference/elements/caption.json` it can replace `stumptown/packaged` with `https://github.com/mdn/stumptown-content/tree/master/content/` and just turn `/caption.json` into `/caption/` so you eventually end up with a URL like:
`https://github.com/mdn/stumptown-content/tree/master/content/html/reference/elements/caption`

However, it would be nice to do more. For example, next to an example it would be cool to have a link directly to the relevant example in stumptown-content. [E.g. this](https://github.com/mdn/stumptown-content/tree/master/content/html/reference/elements/video/examples/multiple-sources-example)

Truth be told, we could technically do a git ref lookup on the `stumptown-content` git submodule and from there figure out which exact version we're using. [E.g. this](https://github.com/mdn/stumptown-content/blob/fcf6289543b2e0d854ddc2ae4a739fb71e39ea30/content/html/reference/elements/video/prose.md#styling-with-css) but I'm not sure how helpful that is. 

So, perhaps we *can* figure out all the relevant GitHub URLs based on the packaged JSON and present these links in the rendered output somehow. In other words, let's try to do something purely in the renderer first and see if it works out.

peterbe:
Forking https://github.com/mdn/stumptown-content/pull/207#issuecomment-548508388 by @ddbeck 

Ideal would be something like this in the built .json files:
```json
    {
      "type": "prose",
      "source": {
          "file": "/Users/peterbe/dev/MOZILLA/MDN/stumptown-content/content/html/reference/elements/video/video.md",
          "line": 1234,
          "column": 23
      },
      "value": {
        "title": "See also",
        "id": "see_also",
[...]

PR #207, opened 2019-10-29
Migrate <acronym>

ddbeck:
OK, this is my first outing to migrate an HTML page and it's pretty rough. I see this PR as a twofer: a migration of a bit of a content and also a set of notes on the migration process itself (potentially leading to improvements to the content or, more likely, issues for improving our tools).

## Content

I used the scraper for this and did a little bit of manual clean up on the source itself (e.g., fixing some unterminated emphasis text). Then I did some things to try to align to the recipe, like converting the example to the `examples/` structure and changing some of the headings to follow the recipe. This had some surprising results (e.g., stumptown-renderer doesn't render the `## Styling` section), but I've left it because it seems correct. Other things are probably not done correctly, so please nit pick the content aggressively.

## Process notes

This was a… taxing process. I spent about 90 minutes trying to make sense of the various error messages to get the content to this point (more PRs to come to help with some of this stuff). Some impressions:

* `build-json`'s frontmatter errors are pretty cryptic sometimes! For example, I got `TypeError: Cannot read property 'element_specific' of undefined
at processMetaIngredient (/Users/ddbeck/Code/Mozilla/stumptown-content/scripts/build-json/build-recipe-page-json.js:32:33)` on a missing bit of frontmatter. I imagine linting will help with some of this stuff, but frontmatter lookups will probably be pretty common—it'd be nice to have some common JS code that looks up some frontmatter and gives a kinder errror when something is missing.
* I put in a blank `contributors.md` file. I don't know how to generate `contributors.md` or whether I should.
* Specifications were a bit surprising. `<acroynm>` is in the HTML 4 spec, which has been superseded. It's not _obviously_ non-standard, but the specs data file says `Never add historical specfications to this allow list (cc: @Elchi3). If that's the way we're doing it, that's fine, but we should consider some guardrails (linting?) for adding spec data that _seems_ like it should OK, but is not.
* The example in this page is really, really basic. The `description.md` boilerplate feels like, well, boilerplate. It'd be nice to have some more affordances for simple examples, such as standalone HTML files or single-file Markdown files with special code fences.
* Testing the page with stumptown-renderer seems like a chore, at least following the README. It seems like it necessarily means building the JSON multiple times (once in my own source, once in the submodule) and it takes a while. This is pretty frustrating while you learn that the renderer currently doesn't seem to render additional prose sections or certain recipe-defined sections (e.g., `## Styling`).

I don't expect any action on this stuff, but I'd like a look over these before I open a bunch of uninformed issues. 😆


wbamberg:
> stumptown-renderer doesn't render the ## Styling section

See https://github.com/mdn/stumptown-content/issues/171.

content/html/reference/elements/acroynm/acronym.md
```diff
@@ -0,0 +1,44 @@
...
+
+Though the purpose of this tag is purely for the convenience of the author, its default styling varies from one browser to another:
+
+-   Some browsers, like Internet Explorer, do not style it differently than a [`<span>`](/en-US/docs/Web/HTML/Element/span) element.
+-   Opera, Firefox, Chrome, and some others add a dotted underline to the content of the element.
+-   A few browsers not only add a dotted underline, but also put it in small caps; to avoid this styling, adding something like [`font-variant`](/en-US/docs/Web/CSS/font-variant)`: none` in the CSS takes care of this case.
+
+It is therefore recommended that web authors either explicitly style this element, or accept some cross-browser variation.
+
+## Browser compatibility
```
wbamberg: This heading and the boilerplate paragraph after it should be removed (and the scraper should be updated to do this for us, I suppose).

content/html/reference/elements/acroynm/acronym.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Short description
+
+The HTML Acronym Element (`<acronym>`) allows authors to clearly indicate a sequence of characters that compose an acronym or abbreviation for a word.
+
+## Overview
+
+This element has been removed in HTML5 and shouldn't be used anymore. Instead web developers should use the [`<abbr>`](/en-US/docs/Web/HTML/Element/abbr) element.
+
+## DOM Interface
```
wbamberg: This is getting outside the scope of "migration", but this is an example of a section that I'd just delete. We will capture the DOM interface as structured content, and noone cares about what happened in Gecko 1.9.2.

content/html/reference/elements/acroynm/acronym.md
```diff
@@ -0,0 +1,44 @@
+---
+recipe: html-element
```
wbamberg: I like having `recipe` first but most of our files have it last. We should consider having a common order (maybe just: the recipe comes first, then items are the same order as in the recipe?).

content/html/reference/elements/acroynm/acronym.md
ddbeck: Opened #210 and removed.

content/html/reference/elements/acroynm/acronym.md
ddbeck: Done.

content/html/reference/elements/acroynm/acronym.md
ddbeck: I like the idea of having the recipe first and deriving the order from the recipe _a lot_. Fixing these automatically could be very amenable to scripting. Like prettier or other formatters, you could run it on save, so you don't even have to think about it, at least as an author.

ddbeck: Opened #212

peterbe:
> Testing the page with stumptown-renderer seems like a chore

:(

It really shouldn't. 

I suspect it's related to how you hook up the "two stumptown-contents". I.e. the `STUMPTOWN_CONTENT_ROOT` env var. 

I have `stumptown-content` in two different places on my hard drive. 

1. `~/dev/MOZILLA/MDN/stumptown-content`
2. `~/dev/MOZILLA/MDN/stumptown-renderer/stumptown`

One is just the git submodule where I don't do branch work, the other is the place I hang most when working on scripts and stuff. 

It *should* work to do this:
In one terminal...
```bash
cd stumptown-renderer
STUMPTOWN_CONTENT_ROOT=/my/preferred/place/for/stumptown-content yarn start
```
In another terminal...
```
open http://localhost:3000/en-US/docs/Web/HTML/Element/video
emacs /my/preferred/place/for/stumptown-content/content/html/reference/element/video/video.md
```
And that should trigger a build with `build-json` for just that one file and it should near instantly refresh your browser.

If this isn't working, we should fix it.

peterbe:
> I don't expect any action on this stuff,

But I do :)

Would you mind filing an issue on the renderer if there's something buggy or wonky with the automatic refresh on edits. 
After all, much of that scripting was writtin rather rushed to demo the concept and since I got it working I haven't really looked back. 
I'm pretty sure it does a poor job letting you know, in the browser, if anything went wrong.

ddbeck:
@peterbe thank you for filling me in on the build stuff! That _is_ much nicer to work with. After I read the README for stumptown-renderer, I thought I needed to have terminals open for all three of `yarn workspace server start`, `yarn workspace client start`, and `yarn cross-env STUMPTOWN_CONTENT_ROOT=path/to/stumptown-content yarn build`. Seems to work great with your method—I'll put it on my to do list to revise the README, as it seems to be a docs problem. 😄

peterbe:
@ddbeck Much of the good and nifty stuff in stumptown-renderer is credited to @wbamberg for  constructively criticizing how it is to work with. 
Let's not underestimate the importance of that tooling story to be **simple** and **user-friendly**. 

By the way, something that's only in planning/as-soon-as-time-allows stages is `docker`. Imagine if the README could just be:

```
To get started, type this:

  git clone --recursive https://github.com/mdn/stumptown-renderer.git
  cd stumptown-renderer
  docker-compose up

Now open http://localhost:3000

If you prefer to run Node and Yarn natively and have more control over everything, consider doingg:

  yarn 
  yarn build
  yarn start

```

peterbe:
Now that I have your attention @ddbeck, one not so crazy idea that we've discussed is to render with "source maps" that link back to the .md files. 

It's actually not unrealistic at all, but if you could know for every `prose` (or `examples`) section, roughly which line in which .md file it belongs you'd be able to do something like double-click on a paragraph or title in the browser when navigating localhost:3000 and it'd immediately open your editor/IDE at that exact point. It might be the kind thing that only works in local development at first but I could even see it as a way to engage people visiting a production site how individual headings or paragraphs link to a URL in GitHub.com as one way to incentivize people to contribute PRs.

ddbeck:
> It's actually not unrealistic at all, but if you could know for every prose (or examples) section, roughly which line in which .md file it belongs you'd be able to do something like double-click on a paragraph or title in the browser when navigating localhost:3000 and it'd immediately open your editor/IDE at that exact point. It might be the kind thing that only works in local development at first but I could even see it as a way to engage people visiting a production site how individual headings or paragraphs link to a URL in GitHub.com as one way to incentivize people to contribute PRs.

Every word of this, yes! I think we could build the JSON body "atoms" with some additional metadata, about file paths and line numbers pretty easily (`unified`'s ASTs can provide line numbers for HTML and Markdown). Maybe this is something we can chat about next week for a bit? (I don't think it's a big thing, but it'd probably be easier to explain what I'd do to build this in 20 minutes rather than typing out a bunch of words.)

peterbe:
This last thread is moved to here: https://github.com/mdn/stumptown-content/issues/77#issuecomment-548513900

PR #208, opened 2019-10-29
Start the linter on a specific directory

ddbeck:
A quick-and-dirty PR that lets you provide a starting directory to the linter, like so:

```shell
$ npm run lint-md -- content/html/reference/elements/abbr
```

This means you can lint only the content you're working on and ignore warnings or errors for other stuff.

### DIFF
--- a/scripts/linter/index.js
+++ b/scripts/linter/index.js
@@ -34,6 +34,10 @@ function main(args) {
 
     const reportedFiles = [];
 
+    if (args.length === 0) {
+        args.push(undefined); // use default starting directory
+    }
+
     for (const arg of args) {
         for (const fp of walkDocs(arg)) {
             const file = vfile.readSync(fp);
--- a/scripts/linter/index.js
+++ b/scripts/linter/index.js
@@ -35,7 +35,7 @@ function main(args) {
     const reportedFiles = [];
 
     if (args.length === 0) {
-        args.push(undefined); // use default starting directory
+        args.push("content");
     }
 
     for (const arg of args) {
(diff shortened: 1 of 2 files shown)


peterbe:
Why the `--`? If you do `npm run lint-md content/html/reference/elements/abbr` instead, won't that `process.args` slice be exactly the same?

scripts/linter/index.js
```diff
@@ -34,10 +34,16 @@ function main() {
 
     const reportedFiles = [];
 
-    for (const fp of walkDocs()) {
-        const file = vfile.readSync(fp);
-        markdownParser().processSync(file);
-        reportedFiles.push(file);
+    if (args.length === 0) {
+        args.push(undefined); // use default starting directory
```
peterbe: I would much rather you edit `walk-docs.js` to not have a default location and move that here to the `main()` function. Then `walk-docs.js` becomes more agnostic we won't need this "hack" here.

peterbe (changes requested):
I think the default `"content"` should be moved to the `main()` function but it's relatively "nitty".

What's missing is that there's no mention of this new feature in the README.

peterbe:
> Yes, it is the same. I'm just in the habit of using `--` with `npm run <whatever> -- <someargs to whatever>` to be explicit that I'm supplying arguments or flags to the script and not `npm` itself. `npm` will eat flags otherwise (e.g., running `npm run some-command --someflag some/path/to/something` means that `--someflag` never reaches `some-command`). I've been bitten by that a few times.
> 
Thank! I'll try to get that into my bloodstream going forwards. 

> I'll fix the default hack.
> 
Thanks!

> The linter is not documented at all. That seems bigger than this change. Is it OK if I open a separate PR for that?

Sure thing. It doesn't have to be lengthy. Just a note or two that it exists. Kinda like the README currently does for the `build-json` script.

peterbe (approved):

PR #209, opened 2019-10-29
Lint for missing frontmatter that will break build-json

ddbeck:
While working on #207, I wrote this to check for `build-json`-breaking frontmatter omissions. It's only got a few right now, and they're hard coded, but I imagine we can build on this. Thankfully, this doesn't introduce any _new_ warnings. 😃

### DIFF
--- a/scripts/linter/plugins/required-frontmatter.js
+++ b/scripts/linter/plugins/required-frontmatter.js
@@ -1,0 +1,45 @@
+const ruleId = "stumptown-linter:frontmatter-required-keys";
+
+/**
+ * If a tree has `tree.data.recipe` (i.e., it's a Markdown doc with a
+ * known-valid recipe), then log messages if it's missing essential frontmatter.
+ */
+function attacher() {
+    return function transformer(tree, file) {
+        if (tree.data && tree.data.recipe) {
+            const { yaml } = tree.children[0].data;
+            const requiredFrontmatter = [];
+
+            // Right now, there's no way to collect the required frontmatter
+            // keys from the recipe itself, so they're hardcoded for now.
+            if (yaml.recipe === "html-element") {
+                requiredFrontmatter.push(
+                    "attributes",
+                    "attributes.global",
+                    "examples"
+                );
+            }
+
+            for (const key of requiredFrontmatter) {
+                // This Rube Goldberg machine will try to look up `a.b.c` and
+                // throw (and log a message) if `a.b.c` is undefined or if any
+                // of `a` or `b` are undefined
+                try {
+                    const obj = key.split(".").reduce((o, i) => o[i], yaml);
+                    if (obj === undefined) {
+                        throw `${key} not found`;
+                    }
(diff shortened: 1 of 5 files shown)


scripts/linter/plugins/required-frontmatter.js
```diff
@@ -0,0 +1,45 @@
...
+ * known-valid recipe), then log messages if it's missing essential frontmatter.
+ */
+function attacher() {
+    return function transformer(tree, file) {
+        if (tree.data && tree.data.recipe) {
+            const { yaml } = tree.children[0].data;
+            const requiredFrontmatter = [];
+
+            // Right now, there's no way to collect the required frontmatter
+            // keys from the recipe itself, so they're hardcoded for now.
```
wbamberg: Can't you look at `meta.` keys that don't have a `?` suffix? This will fail because we don't yet understand `info_box`, but it might be better to do this and comment `info_box` out of the recipe. Or is there some other reason this won't work?

ddbeck: This would partially work. `examples` would be fixed by checking the recipe. But `attributes` is a weird one: it requires some nested keys that aren't in the recipe. `global` is conventional, though best as I can tell it's not used to build any JSON yet; `element_specific` is used in `build-json` but the only place it's described is in `scripts/build-json/build-recipe-page-json.js`.

There's a little bit of tension here in that I want to check that some frontmatter _exists_, but I'm not ready to build out complete checks for well-formed frontmatter (mostly because I don't think we actually know what well formed is yet).

ddbeck: OK, that sounds reasonable. I've changed it to check the recipe for top-level required frontmatter and ignore the nested keys. We can tackle the finer details (e.g., checking for valid spec URLs) in other PRs. Want to take another look?

scripts/linter/plugins/required-frontmatter.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Extract the keys of required top-level frontmatter from a recipe.
+ */
+function required(recipe) {
+    return recipe.body
+        .map(entry => {
+            if (typeof entry === "string") {
+                return entry;
+            } else {
+                return Object.keys(entry)[0];
```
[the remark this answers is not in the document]
wbamberg: For HTML elements we've removed the nested pieces, so the recipe is like:

```
...
- prose.*
- meta.examples
- meta.info_box
- meta.specifications
- meta.browser_compatibility
...
```

I think we should do the same for CSS properties. (I haven't done any maintaining of the CSS properties while I've been wrestling only with HTML documentation, so it has fallen behind. Which was fine, except we are now linting them).

wbamberg (approved):
Thank you @ddbeck , this looks good and works well.

It does now introduce new errors, for the CSS properties. I'm not sure whether we should fix them, or just delete the CSS property pages, since we're not focusing on them at the moment. Probably the latter.

PR #222, opened 2019-11-01
Remove BCD section text during scraping

### LINKED ISSUE #210: Scraper should remove "Browser compatibility" section and text
> This heading and the boilerplate paragraph after it should be removed (and the scraper should be updated to do this for us, I suppose).

_Originally posted by @wbamberg in https://github.com/mdn/stumptown-content/pull/207_

ddbeck:
I tired my hand at modifying the scraper to remove BCD sections from the final prose Markdown.

`scrape-mdn.js` has not been prettier-ed yet, so my first commit here does that. You might find this easier to review if you use GitHub's diff commit picker to see only the meaningful changes in the second commit (0737b6e).

Fixes #210.

### DIFF
--- a/scripts/scraper/clean-html.js
+++ b/scripts/scraper/clean-html.js
@@ -19,7 +19,27 @@ function removeHiddenDivs(dom) {
   }
 }
 
+function removeBrowserCompatibility(dom) {
+  // Note: This assumes that hidden content has been removed and macros have been processed out of dom.
+
+  const compatHeading = dom.window.document.querySelector(
+    "#Browser_compatibility"
+  );
+
+  // Make sure that there's no content under the compat heading by making sure
+  // the next elmeent after the compat heading is another heading.
+  const nextElement = compatHeading.nextElementSibling;
+  if (nextElement !== null && nextElement.tagName !== "H2") {
+    console.error("There's unexpected content in the BCD section.");
+    console.error(nextElement, nextElement.tagName, nextElement.innerHTML);
+    process.exit(1);
+  }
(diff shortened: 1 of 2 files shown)


ddbeck:
Hey @wbamberg, I've added a check for non-BCD-table content in the compat section. Is this more along the lines of what you were thinking of?

ddbeck:
Will suggested not assuming that hidden content should be dropped _everywhere_.

ddbeck:
@wbamberg OK, this version wraps the compat section body in its own div, removes hidden elements, then sees if there's anything left. If there is, it fails (and prints out the unexpected content). If there isn't, then we remove the compat section body wrapper and heading and we prevent the heading from reaching the Markdown format.

scripts/scraper/clean-html.js
```diff
@@ -12,7 +12,44 @@ function removeNode(dom, selector) {
...
+
+  if (sectionWrapper.childNodes.length > 0) {
+    console.error("There's unexpected content in the BCD section.");
+    console.error(sectionWrapper.innerHTML);
+    process.exit(1);
+  }
+
+  // Finally, remove the heading
+  heading.parentNode.removeChild(heading);
+  sectionWrapper.parentNode.removeChild(sectionWrapper);
```
Elchi3: Can these two lines call `remove`? 
```js
heading.remove();
sectionWrapper.remove();
```

ddbeck:
Good call on the DOM methods, Florian. I've pushed some updates. Thank you!

PR #223, opened 2019-11-01
Lint for specifications section

ddbeck:
As a courtesy to authors using the scraper, this change will caution you against leaving in a `## Specifications` section (since you should use the `specifications` frontmatter instead).


wbamberg (approved):
Thanks @ddbeck !

PR #227, opened 2019-11-06
Linter: fix bug visiting headings repeatedly

ddbeck:
The return value of the visitor function for [unist-util-visit](https://github.com/syntax-tree/unist-util-visit#next--visitornode-index-parent) is used to continue (`true`), break (`false`), or change the visit (with an index or node). I was accidentally returning the length of an array, which caused headings to be visited multiple times. This PR fixes that problem.


wbamberg (approved):
Thanks @ddbeck !

PR #229, opened 2019-11-06
Prettier YAML

ddbeck:
This PR turns on prettier for YAML files in the repo, for consistent YAML indentation and the like.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,7 @@ branches:
 
 script:
   - npm run ci-checks
-  - npm run lint-js
-  - npm run lint-md
-  - npm run spell-md
+  - npm test
   # Note, in a near future we're going to want to build-json for *everything*
   # and not just the 'html' section.
   - npm run build-json html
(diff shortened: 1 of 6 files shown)


ddbeck:
This will conflict with #228 in some way (requiring a rebase here or there, whatever the merge order).

ddbeck:
Rebased.

peterbe (approved):
The only flaw I see is that this is running under `npm run lint-js` (in Travis) which is now poorly named. We might Prettier check other stuff some day such as the CSS in the examples. 
Perhaps we just rename that script in package.json to `lint-code` :)

ddbeck:
> Perhaps we just rename that script in package.json to `lint-code` :)

Good idea, @peterbe! Done.

peterbe (approved):

ddbeck:
Fixed the build failure. Didn't catch dotfiles in my search for `npm run lint-js`

PR #230, opened 2019-11-07
Add BigInt class to stumptown

wbamberg:
This PR adds [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) to stumptown.

Specifically it adds:
* 2 new recipes, "javascript-class" and "javascript-method"
* content for the `BigInt` class and its methods
* code to understand the new ingredients `static_methods` and `instance_methods`

This will break the renderer, since the renderer has to be taught about the new ingredients. But it should be really simple to actually render them, because they are just `link_list` things, that the renderer already knows how to render.


ddbeck:
Thanks, Will. I'm happy with this being merged at your discretion. One note:

> This will break the renderer

If you wanted to add a file containing the hash of the last known-good renderer in this (or another) PR, that would be a welcome, but optional, development.

peterbe:
Breaking the renderer is actually “harmless” because of the nature of the direction of the sub module nature. 
That versioning, in the other direction, will matter when we add building-the-renderer s part of this repos CI.

wbamberg:
> If you wanted to add a file containing the hash of the last known-good renderer in this (or another) PR, that would be a welcome, but optional, development.

Sorry, I don't understand. After this PR is merged (and before the renderer is updated) there isn't a version of the renderer that works with this version of content. Or do you mean the other way round, that we have a file in stumptown-renderer containing a hash of the most recent version of stumptown-content that this version of the renderer understands?

ddbeck:
Yeah, it's not terribly useful in either direction, at least not yet. Ignore that request for now. :-)

PR #231, opened 2019-11-08
Add a linter for specifications

Elchi3:
As said (and specified) in https://github.com/mdn/stumptown-content/issues/126, here's a linter for the `specifications` front-matter.

### DIFF
--- a/scripts/linter/plugins/valid-specs.js
+++ b/scripts/linter/plugins/valid-specs.js
@@ -1,0 +1,53 @@
+const visit = require("unist-util-visit");
+const yaml = require("js-yaml");
+const fs = require('fs');
+const path = require('path');
+
+const ROOT = path.join(__dirname, "..", "..", "..");
+const dataDir = path.resolve(ROOT, 'content', 'data');
+const specMap = yaml.safeLoad(fs.readFileSync(path.join(dataDir, 'specifications.yaml'), 'utf8'));
+
+const ruleId = "stumptown-linter:valid-specifications";
+const specURLRegex = new RegExp("^http(s)?://[^#]+#.+");
+
+/**
+ * Check the front-matter for valid `specifications`
+ */
+function attacher() {
+    return function transformer(tree, file) {
+        visit(tree, "yaml", node => {
+            let specs = node.data.yaml.specifications;
+            // specifications are optional
+            if (specs) {
+                // specifications can be single url or array of urls
+                if (!Array.isArray(specs)) {
+                    specs = [specs];
+                }
+                specs.forEach(spec => {
+                    // "non-standard" is a special string indicating that there is no spec. Allow this one.
+                    if (spec === 'non-standard') {
+                      return;
+                    }
+                    // Any other strings should be URLs and provide a deep link to a spec section.
+                    if (!specURLRegex.test(spec)) {
+                        const message = file.message(`"${spec}" is not a valid specification link (anchored deep link required).`, node, ruleId);
+                        message.fatal = true;
+                    }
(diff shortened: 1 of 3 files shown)


scripts/linter/plugins/valid-specs.js
```diff
@@ -0,0 +1,53 @@
...
+                    if (spec === 'non-standard') {
+                      return;
+                    }
+                    // Any other strings should be URLs and provide a deep link to a spec section.
+                    if (!specURLRegex.test(spec)) {
+                        const message = file.message(`"${spec}" is not a valid specification link (anchored deep link required).`, node, ruleId);
+                        message.fatal = true;
+                    }
+                    // Spec URLs should be allow-listed, so that we don't list any outdated specs
+                    let found = false;
```
ddbeck: I found the `let found … if (!found)` sequence a bit hard to follow. What do you think of something like this, instead?

```js
const allowedSpecs = Object.keys(specMap);
if (!allowedSpecs.some(key => spec.includes(key)) {
  const message = file.message(`Domain for "${spec}" not found in data/specifications.yaml`, node, ruleId);
  message.fatal = true;
}
```

scripts/linter/plugins/valid-specs.js
```diff
@@ -0,0 +1,53 @@
...
+const ruleId = "stumptown-linter:valid-specifications";
+const specURLRegex = new RegExp("^http(s)?://[^#]+#.+");
+
+/**
+ * Check the front-matter for valid `specifications`
+ */
+function attacher() {
+    return function transformer(tree, file) {
+        visit(tree, "yaml", node => {
+            let specs = node.data.yaml.specifications;
```
ddbeck:
```suggestion
            const specs = node.data.yaml.specifications;
```

Elchi3: I'm actually modifying `specs` to be always an array in line 24, so I think this should stay `let`.

scripts/linter/plugins/valid-specs.js
Elchi3: This is great! Thank you! :)

scripts/linter/plugins/valid-specs.js
ddbeck: Oh I missed that `specs` is reassigned below. Disregard!

Elchi3:
Rebased :)

ddbeck:
Oops, that broke against part of the fr language content. Localization surprises

https://travis-ci.org/mdn/stumptown-content/jobs/612402273#L576

wbamberg:
@SphinxKnight , the linter complains that [content/fr/html/HTML.md](https://raw.githubusercontent.com/mdn/stumptown-content/master/content/fr/html/HTML.md) contains:

    specifications: https://html.spec.whatwg.org/multipage/

...where it expects a spec link to contain a fragment. The corresponding file at [content/en-US/html/HTML.md](https://raw.githubusercontent.com/mdn/stumptown-content/master/content/en-US/html/HTML.md) contains:

    specifications: https://html.spec.whatwg.org/multipage/introduction.html#abstract

Is there a reason these should be different?

But also, I wonder if we should relax the check that a spec URL must have a fragment. https://html.spec.whatwg.org/multipage/ seems like a more useful link in this case than https://html.spec.whatwg.org/multipage/introduction.html#abstract.

SphinxKnight:
I'm opening a PR to change the fr version as well :)

Elchi3:
> Oops, that broke against part of the fr language content. Localization surprises

Sorry about that. I should have rebased this again.

SphinxKnight:
*mumble grumpily in French* :)

PR #234, opened 2019-11-08
Adding a locales objects to the packaged JSON for localized docs "linkability"/"discoverability"

SphinxKnight:
__This introduces a "schema" change to the final JSON file__

The intent of this change is to be able to know the URL of the localized versions of a given document ("I know that the en-US version's URL is /en-US/Learn/Learning_HTML/How_to_annotate but where is the French version if any?").

In order to solve this, I introduce a `locales` property which is an array of `{locale, mdn_url}` objects where `locale` is the short code ("en-US", "fr", "ja", etc.) and `mdn_url` is derived from the attribute in the document Markdown page.

Before (`packaged/en-US/html/reference/elements/abbr.json` for example): 
```
{
  "title": "<abbr>: The Abbreviation element",
  "mdn_url": "/en-US/docs/Web/HTML/Element/abbr",
  "related_content": [
    {
```

After:
```
{
  "title": "<abbr>: The Abbreviation element",
  "mdn_url": "/en-US/docs/Web/HTML/Element/abbr",
  "locales": [
    {
      "locale": "en-US",
      "mdn_url": "/en-US/docs/Web/HTML/Element/abbr"
    },
    {
      "locale": "fr",
      "mdn_url": "/fr/docs/Web/HTML/Element/abbr"
    }
  ],
  "related_content": [
    {
...
```
Those commits are based on the ones from https://github.com/mdn/stumptown-content/pull/233 but I might rebase them (if needed and if I manage to do so :) ). For this feature, only the following commits are relevant: 
* https://github.com/mdn/stumptown-content/commit/6fdcad44422b56a5ebb1b0831577377b407c8878
* https://github.com/mdn/stumptown-content/pull/234/commits/8dd0fa07d91c3a6fe45b68746604d1e6e713b071

Having such a property allows for "consumers" (as stumptown-renderer) to link between each locale when a localized version exists (e.g. having a "Languages" dropdown in the UI).

N.B.2 Though I rendered things with stumptown-renderer without any issue after those changes, as this introduces a change in the schema, poking both @Elchi3 & @wbamberg for content and @peterbe for stumptown-renderer.


SphinxKnight:
Aparts from conflicts (that I think I can fix), is there anything blocking this PR from merging/triaging?

wbamberg:
Sorry @SphinxKnight for not looking at this yet. I have been meaning to, but I have struggled with understanding the whole problem of localized versions of pages. So I will review this, but think it would be really great to get some feedback from the rest of the stumptown crew as well.

peterbe:
@SphinxKnight We are very appreciative of your work here but sorry that we're so slow to take it in. 
@escattone and I have been working hard on this very same problem in Yari (remember, Yari is Kuma done with git instead of MySQL, and a Node CLI instead of a Django web server). In Yari, we won't have Markdown (unfortunately) but instead each document will be its own folder and will have the following:

* `index.html` (the stuff you see when you click "Source" in the Wiki `$edit`)
* `index.yaml` and the contents of this file will look like this:

```yaml
title: Microdata
slug: Web/HTML/Microdata
```
or, if it's a translation
```yaml
title: Microdonnées
slug: Web/HTML/Microdonnées
translationOf: Web/HTML/Microdata
```
What's implicit is:

1. The `translationOf` slug **always** refers to the en-us version. 
2. The locale of the document is implied by its location on disk

In the Yari builder because it first goes through all `index.yaml` files once, before it starts actually building the final HTML pages, it knows all relationships between all documents. From this, produces an intermediate structure that looks something like this:

```json
{
  "doc": {
    "sidebarHTML": "<ol>...",
    "title": "CacheStorage",
    "mdn_url": "/en-US/docs/Web/API/CacheStorage",
    "body": [  ],
    "other_translations": [
      {
        "locale": "fr",
        "slug": "Web/API/CacheStorage"
      },
      {
        "locale": "ru",
        "slug": "Web/API/CacheStorage"
      },
      {
        "locale": "zh-CN",
        "slug": "Web/API/CacheStorage"
      }
    ],
```

And if it's a translated document, it looks like this:

```json
{
  "doc": {
    "sidebarHTML": "<ol>...",
    "title": "CacheStorage",
    "mdn_url": "/fr/docs/Web/API/CacheStorage",
    "parent": {
      "slug": "Web/API/CacheStorage",
      "locale": "en-US"
    },
    "body": [ ],

    "other_translations": [
      {
        "locale": "fr",
        "slug": "Web/API/CacheStorage"
      },
      {
        "locale": "ru",
        "slug": "Web/API/CacheStorage"
      },
      {
        "locale": "zh-CN",
        "slug": "Web/API/CacheStorage"
      }
    ],
```

Lastly, what the client-side code does it looks at the structure for if the document has a `.other_translations` and a `.parent`. Then, to make the little language choice drop down it basically takes the list of `[parent] + other_translations`. 

What's cool about this is 

* Writers have absolutely minimal in their YAML. But a linter *will* check that every non-en-us file has a `translationOf` and that slug needs to be existent or else it will break the build.
* You can mess with the slugs as much as you want in any non-en-US document. 
* If you edit the slug in the en-US one you better be prepared to edit the `translationOf` value in all the translations. (linting in CI will prevent corruption)
* From this we can compute, on a en-US document, links to all other locales. And one a non-en-US document, we can do the same but with English being the first choice. 
* Subtle, but important, it doesn't matter what the **name of the folder** is. The "locale folder name" matters so you can't rename it from "sv-se" to "sv-sw" without things breaking. 

So what I'd like you to consider is the one and only key of `translationOf`. (I'm personally not a huge fan of the use of mixed cased since this is yaml, not JavaScript). 

It's very possible that I don't understand recipes and related_content in stumptown-content  but what I'm suggesting is something like this:

```diff
---
title: "<abbr> : L'élément pour les abréviation"
short_title: <abbr>
mdn_url: /fr/docs/Web/HTML/Element/abbr
+translation_of: /en-US/docs/Web/HTML/Element/abbr
specifications: https://html.spec.whatwg.org/multipage/semantics.html#the-abbr-element
tags:
    group: Inline text semantics 
api: HTMLElement
permitted_aria_roles:
    - any
tag_omission: none
interactive_example:
    url: https://interactive-examples.mdn.mozilla.net/pages/tabbed/abbr.html
    height: html-short
browser_compatibility: html.elements.abbr
examples:
    - examples/baliser-une-abréviation
    - examples/mettre-en-forme-une-abréviation
    - examples/fournir-une-forme-étendue
    - examples/définir-une-abréviation
attributes:
    global: /content/fr/html/global_attributes
recipe: html-element
---
```


The bigger ticket to discuss is; who should compute these things. In Yari or in `stumptown-content`'s own `build-json` Node scripts. Either

* `build-json` mimics what Yari does when it makes those `packaged/**/*.json` files
* `build-json` doesn't mess with this at all and Yari will bundle this 100% (my favorite).

escattone:
Just wanted to add my support for @peterbe's really nicely done comment immediately above, as well as say that [Hugo does something similar](https://gohugo.io/content-management/multilingual/#bypassing-default-linking) with their `translationKey`.

SphinxKnight:
It's been a year and a lot happened. I'm closing this one.

PR #247, opened 2019-11-15
Fix regressions and errors in JS support

wbamberg:
The fix for https://github.com/mdn/stumptown-content/pull/243 was not complete: the locale support work also meant that recipes now refer to related_content using a path relative to the current locale.

Also, my PR to add JS support regressed interactive examples.

This PR is intended to fix both of these problems, so JS pages should more or less work.

I've tested it using a version of the renderer modified to render the JS pages.

### DIFF
--- a/recipes/html-input-element.yaml
+++ b/recipes/html-input-element.yaml
@@ -1,4 +1,4 @@
-related_content: /related_content/html.yaml
+related_content: related_content/html.yaml
 data:
   - title
   - short_title?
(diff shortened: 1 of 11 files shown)


ddbeck (approved):
This seems fine to me. I wish there was a way to indicate that it's locale relative (rather than relative to the recipes themselves), but I can't think of anything that wouldn't be more confusing, so I'm happy with this. 👍

PR #248, opened 2019-11-15
Fix spec link that fails the build

ddbeck:
This slipped through on https://github.com/mdn/stumptown-content/pull/231#issuecomment-554377417.

### DIFF
--- a/content/fr/html/HTML.md
+++ b/content/fr/html/HTML.md
@@ -1,7 +1,7 @@
 ---
 title: "HTML: Hypertext Markup Language"
 mdn_url: /fr/docs/Web/HTML
-specifications: https://html.spec.whatwg.org/multipage/
+specifications: https://html.spec.whatwg.org/multipage/#table-of-contents
 recipe: landing-page
 related_content: /related_content/html.yaml
 link_lists:


wbamberg:
Oh, we crossed over on this :)

-> https://github.com/mdn/stumptown-content/pull/231#issuecomment-554550071

The corresponding file at [content/en-US/html/HTML.md](https://raw.githubusercontent.com/mdn/stumptown-content/master/content/en-US/html/HTML.md) contains:

    specifications: https://html.spec.whatwg.org/multipage/introduction.html#abstract

...so this PR means the link diverges between en-US and fr, which I assume we don't want[1]. So this PR could either:

* update only the fr link to match the en-US version
* update both links to be https://html.spec.whatwg.org/multipage/#table-of-contents

But also, for "HTML.md", the best link seems to be just https://html.spec.whatwg.org/multipage/, so I worry that we are making the content less good to please the linter, and we might be better off to relax the linter's requirement for links to contain fragments.

[1] which itself is a can of worms: should we/can we help ensure that localized content doesn't diverge in places where it shouldn't?

ddbeck:
First, #249 fixes this more correctly; I'm going to close this in favor of @SphinxKnight's PR. I was just trying to suppress the red x and didn't really think this through.

To address your points, @wbamberg:

> But also, for "HTML.md", the best link seems to be just https://html.spec.whatwg.org/multipage/, so I worry that we are making the content less good to please the linter, and we might be better off to relax the linter's requirement for links to contain fragments.

That's a good point. I'm going to turn this into an issue. I'd especially like Florian's take on this.

> [1] which itself is a can of worms: should we/can we help ensure that localized content doesn't diverge in places where it shouldn't?

I don't want to think about this.

The serious answer is: yes. Firstly, I'd hope to make it so you don't have to duplicate content like this at all (e.g., unless you supply a localized specification link—an fr equivalent, in this case—build-json uses the frontmatter from the corresponding en-US page). You can't diverge if you never enter it in the first place.

But to be honest, I'm not sure how to articulate this as an issue. It seems like a high-level design goal. Maybe the right place to start is to see if there are other docs-as-code projects that have dealt with this kind of thing?

wbamberg:
> Firstly, I'd hope to make it so you don't have to duplicate content like this at all (e.g., unless you supply a localized specification link—an fr equivalent, in this case—build-json uses the frontmatter from the corresponding en-US page). You can't diverge if you never enter it in the first place.

Yes, it would be quite simple to change build-json to:
- look in this MD file for `data` properties
- if they are not found in this one, look in the en-US version of this MD file

It would be even better to say front matter properties just aren't localizable. So far at least I think the only ones that need to be are `title`, `short_title`, and `mdn_url`. We might consider explicitly distinguishing localizable properties, so we can lint for this and build-json can only look in en-US for the non-localizable ones.

Elchi3:
> That's a good point. I'm going to turn this into an issue. I'd especially like Florian's take on this.

Commented in the issue. Thanks for filing it, Daniel!

> [1] which itself is a can of worms: should we/can we help ensure that localized content doesn't diverge in places where it shouldn't?

I propose we move this to a new issue as well, but I think `specifications` shouldn't be localizable. I think it is great that we could control things like this in the new world. It is opportunity for less outdated localized docs imo.

ddbeck:
> I propose we move this to a new issue as well, but I think `specifications` shouldn't be localizable

Opened as #252.

PR #262, opened 2019-12-13
Fix some errors from the sitemap test

wbamberg:
This fixes `mdn_url` in the CSS properties, which should resolve 3 of the 4 CI errors.


wbamberg:
But there's a new error because https://wiki.developer.mozilla.org//en-US/docs/Web/HTML/CORS_settings_attributes has recently been redirected to https://wiki.developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin.

I can (and will) just delete it from stumptown, but I guess it's a bit worrying that changes in the Wiki can now break CI in stumptown.

PR #271, opened 2020-01-06
Normalize macro names in linting

### LINKED ISSUE #270: Normalize macro names in html-require-macros
The [html-require-macros](https://github.com/mdn/stumptown-content/blob/master/scripts/scraper-ng/rules/html-require-macros.js) rule checks that a document contains all the macro calls listed in the `requiredMacros` array.

The comparison is a direct string compare, but macro calls are case-insensitive. So this will generate false positives in pages like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator, where the macro is named `compat` but the `requiredMacros` array contains `Compat`.

ddbeck:
This is my attempt to handle macro name case better. I've distinguished a macro's _name_ (lower case) from the _call_ used in the linted page (whatever case the author of the page actually used). Hopefully this approach will let the linter messages accurately reflect the source, while matching macro names case insensitively.

Fixes #270.

### DIFF
--- a/scripts/scraper-ng/rules/html-require-macros.js
+++ b/scripts/scraper-ng/rules/html-require-macros.js
@@ -1,6 +1,8 @@
 const rule = require("unified-lint-rule");
 const visit = require("unist-util-visit");
 
+const normalizeMacroName = require("../normalize-macro-name");
+
 /**
  * Require one or more named macros.
  */
@@ -10,9 +12,12 @@ function requireMacros(tree, file, required = []) {
     node =>
       node.type === "text" &&
       node.data &&
-      required.includes(node.data.macroName),
+      required.map(normalizeMacroName).includes(node.data.macroName),
     node => {
-      required = required.filter(macro => node.data.macroName !== macro);
+      required = required.filter(
+        macro => node.data.macroName !== normalizeMacroName(macro)
+      );
+
       if (!required.length) {
         return visit.EXIT;
       }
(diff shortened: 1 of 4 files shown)


scripts/scraper-ng/rules/html-require-macros.js
```diff
@@ -10,9 +12,12 @@ function requireMacros(tree, file, required = []) {
     node =>
       node.type === "text" &&
       node.data &&
-      required.includes(node.data.macroName),
+      required.map(normalizeMacroName).includes(node.data.macroName),
```
peterbe: I have no idea what I'm talking about but... Why isn't this
```suggestion
      required.map(normalizeMacroName).includes(normalizeMacroName(node.data.macroName)),
```

ddbeck: `node.data.macroName` is already normalized (see [this bit](https://github.com/mdn/stumptown-content/pull/271/files#diff-e403f425e4ea2799c526b9b5ad22f009R17)).

As for `required.map(normalizeMacroName)`, I did that to preserve the original capitalization in the configuration (see `preset.js`). This way the linting messages can be a bit prettier (e.g., imagine `SomeComplexMacroName required but not found` versus `somecomplexmacroname required but not found`).

peterbe: Got it. I guess you'd have to read the whole code to understand that `node.data` had already gone through the normalization ringer. :)

Also, mayor bonus for being able to produce an error message based on the input rather then the normalized comparison. 
Like those http libs where you can query for an header case-insensitively and it throws and error if it's not present and the error mentions it like the coder mentioned it.

ddbeck: Yeah, I very nearly went with the uglier/surprising output, but thought I should write the code I want to see in the world 😆

Issue #242, opened 2019-11-13
More compact format for examples


wbamberg:
Currently examples each live in their own directory, that contains:
* a "description.md" optionally containing some prose description plus front matter
* a separate file for each .js, .css, .html piece

For many examples this results in a lot of files, many of which contain a line or two. We talked in London about changing to a more compact format, like:

* one MD file per example, containing:
    * front matter
    * a prose section
    * one fenced code section for each language piece (plus perhaps one for hidden pieces)

e.g.

```
    ---
    title: Picking a color
    height: 317
    ---
    Let's look at a simple example, in which the user can choose a color. As the user adjusts the color, the border around the example changes to reflect the new color. After finishing up and picking the final color, the color picker's value is displayed.
    
    ```html
    <div id="box">
      <label for="colorPicker">Border color:</label>
      <input type="color" value="#8888ff" id="colorPicker">
      <p id="output"></p>
    </div>
    ```
    
    ```css hidden
    #box {
      width: 500px;
      height: 200px;
      border: 2px solid rgb(245, 220, 225);
      padding: 4px 6px;
      font: 16px "Lucida Grande", "Helvetica", "Arial", "sans-serif"
    }
    ```
    
    ```js
    let colorPicker = document.getElementById("colorPicker");
    let box = document.getElementById("box");
    let output = document.getElementById("output");
    
    box.style.borderColor = colorPicker.value;
    
    colorPicker.addEventListener("input", function(event) {
      box.style.borderColor = event.target.value;
    }, false);
    
    colorPicker.addEventListener("change", function(event) {
      output.innerText = "Color set to " + colorPicker.value + ".";
    }, false);
    ```
```

@Elchi3 , @ddbeck , thoughts?

********

## Acceptance criteria

Examples format is agreed and implemented, and all stumptown content is updated with the new format.

peterbe:
I was curious how this'd look like in `code`.

Previewing...
<img width="924" alt="Screen Shot 2019-11-13 at 2 46 29 PM" src="https://user-images.githubusercontent.com/26739/68798451-7208f380-0624-11ea-92fe-81e891dd4227.png">

peterbe:
And the extra ` hidden` doesn't seem to break here in GitHub commenting:
<img width="835" alt="Screen Shot 2019-11-13 at 2 48 59 PM" src="https://user-images.githubusercontent.com/26739/68798687-ce6c1300-0624-11ea-8e0b-803174f5e3d1.png">

[...]

PR #281, opened 2020-01-13
Remove old content

wbamberg:
This PR removes all but three of the HTML element pages and all but one of the CSS property pages.

It's clear that the content already imported to stumptown-content is stale and would need to be re-migrated anyway, and that we are now intending to automate this.

At the same time, if we want to implement changes like [More compact format for examples](https://github.com/mdn/stumptown-content/issues/242) we would have to update hundreds of files, which just doesn't seem like a good use of our time.

I've kept a few files in here, that we can use to demonstrate/test features as we implement them:

* one CSS property, `transform`.
* three HTML elements, `article`, `body`, `video`. I've tried to choose ones that have different features.
* the JS `BigInt` pages.
* all the guide pages, just because they are less likely to be a maintenance burden.

We could probably cull some more, but for now this seems like a small enough set that maintenance won't be too time-consuming.


Elchi3 (changes requested):
This makes sense to me, Will.
Since we have a recipe for html-input, can we keep an input type for demonstration/testing?

wbamberg:
Thanks for the review! I kept `input` as well, because `input-button` currently refers to some attributes under there.

Elchi3 (approved):
Thanks @wbamberg :+1: We can always dig out more content from version control, so I'm r+ on this one.

PR #284, opened 2020-01-14
More compact format for examples

wbamberg:
This PR attempts to implement the new example format discussed in https://github.com/mdn/stumptown-content/issues/242 . I haven't tangled with the metadata changes, since that seemed to be controversial.

The main focus of the PR is [compose-examples.js](https://github.com/mdn/stumptown-content/compare/master...wbamberg:compact-examples?expand=1#diff-f6fa87e359f30b84cdaf24f29bb5df8f), which is changed to understand the new format. It just assumes that the file is in the right format: this is in line with the idea AIUI that we will separate linting from building.

I've also migrated three pages to show how it looks:

* `<article>`: this is a really simple non-live sample
* `<input type="button">`: this has some live samples with multiple sources
* `<video>`: this has some live samples with single sources

I've manually verified that the JSON output has no significant differences from the old output.

It does have some insignificant ones. (I assume that because we round-trip through JSDOM), the characters used to escape HTML tags are different, so:

`<p>This example demonstrates the use of <code>&#x3C;article></code> in` 

in the old JSON becomes

`<p>This example demonstrates the use of <code>&lt;article&gt;</code> in`

in the new JSON. But I think this is OK.

I've also tested using stumptown-renderer that these examples still look and behave the same.

I've not yet migrated all the content so am marking this Draft, and CI fails. I will migrate all the other content in this PR if the code change looks OK. But it seemed worthwhile getting some feedback first, before I change everything.


ddbeck:
This looks like the right idea to me. 👍

wbamberg:
Thanks @ddbeck . I've updated all examples to use this approach and tested it using stumptown-renderer. Sorry for the lots of changed files, but most of them are deletions :).

ddbeck:
@wbamberg I'm sorry, I only just now noticed that this was no longer a draft PR. I'll review (and, optimistically, merge) this before our meeting on Wednesday.

content/en-US/html/reference/elements/input/input.md
```diff
@@ -20,9 +20,9 @@ attributes:
         - ../input/attributes/value.md
     global: /content/html/global_attributes
 examples:
-    - examples/Examples_overview
-    - examples/Styling_input_elements
-    - examples/Custom_error_messages
+    - examples/Examples_overview.md
```
ddbeck: A funny thing here: `Examples_overview.md` doesn't actually contain an example. Should this exist? It seems to me that an example that doesn't contain any code blocks is dubious. Or is this a linting issue?

ddbeck (approved):
OK, that turned out to be easier than to review than I expected. Way to go! I'm really excited about what this change represents and how it's going to improve the wiki.

That said, I did call attention to one oddity I discovered along the way. I don't have a strong feeling about it, but I thought I'd be remiss if I didn't mention it. If you think it's fine, you're free to merge as-is.

content/en-US/html/reference/elements/input/input.md
wbamberg: Yes, I think you're right.

This is partly about how much flexibility we should give to writers. In this case they want to just point to another place people can find examples, and this is an easy way to provide that.

But anyone trying to consume these examples *as examples*, rather than as bits of a page, is going to be confused by this. So I agree that we should not do this, and a principle for designing linting rules should be that the examples should contain examples :), even if they are not always executable.

I'm not sure how we should address this desire for flexibility. Perhaps it's not that important. Perhaps writers should always include actual examples (for example, here, instead of vaguely gesturing at some other pages, this page could include some actual examples living under, say, /input/button). Or perhaps we should explicitly represent some "preamble" prose that can be added at the start of an **Examples** section, to give writers that flexibility.

But let's not add more stuff until we're sure we need it. For now I've removed this "example" from the content (actually, this page has been completely rewritten and now contains no examples at all!).

wbamberg:
Thanks for the quick review Daniel! I do think this is a much nicer format for examples, so I'm happy to get it in.

PR #288, opened 2020-01-17
Explore matching pages to recipes

ddbeck:
This is an initial exploration of matching tags for each page to a stumptown recipe, for https://github.com/mdn/sprints/issues/2671.

In the interest of keeping PRs small, I'm opening this PR now even though you can't really judge yet how effective this is for the recipe types specifically. I'm hoping to get feedback on whether this is a good approach to structuring recipe identification; I don't expect this to be the final word on how specific recipe types are matched. 

An important follow up to this PR is going to be a report, with many pages showing how the script matches recipes to pages, and what a manual match would do. I think that data will give us helpful information about what steps we need to take on the wiki and in this code base.

This PR does two big things:

* adds a `kuma-metadata` plugin that fetches `$json` data for pages and attaches some it to the vfile
* adds an `annotate-recipe` that uses the tags list from `kuma-metadata`.

With my earlier caveats in mind, I'm open to any feedback. Some questions I've already asked myself, without good answers yet:

* `kuma-metadata`: Is it a bad name or a terrible name?
* `kuma-metadata`: Should I just attach _all_ the data from `$json` to the vfile?
* `annotate-recipe` warns about missing recipes. Should this be its own check in `/scripts/scraper-ng/rules/` (e.g,. attempt to identify a recipe is one module, but _requiring_ a recipe happens elsewhere)? If not, does this imply that `rules` and `plugins` should just be lumped together?
* `annotate-recipe` goes with the first recipe that matches, mostly because it was easy to write. Should recipe matches be required to be unique (i.e., a page should match one and only recipe?) or should we just concern ourselves with a best effort, by ranking recipe matchers by specificity?

I'm going to flag @wbamberg for review, but tagging @Elchi3 as an FYI. Thanks for taking a look!

### DIFF
--- a/scripts/scraper-ng/plugins/annotate-recipe.js
+++ b/scripts/scraper-ng/plugins/annotate-recipe.js
@@ -69,24 +69,13 @@ function recipe(name) {
 }
 
 /**
- * Check if some tags contains some other tags.
- *
- * @param {Set<String>} actualTags
- * @param {Set<String} expectedTags
- * @returns {Boolean} `true` or `false`
- */
-function hasAll(actualTags, expectedTags) {
-  return isSuperset(actualTags, expectedTags);
-}
-
-/**
- * Check if a set is a subset of another set.
+ * Check if a set of tags contains some other set of tags.
  *
  * @param {Set} set
  * @param {Set} subset
  * @returns {Boolean} `true` or `false`
  */
-function isSuperset(set, subset) {
+function hasAll(set, subset) {
   for (let elem of subset) {
     if (!set.has(elem)) {
       return false;
--- a/scripts/scraper-ng/plugins/annotate-recipe.js
+++ b/scripts/scraper-ng/plugins/annotate-recipe.js
@@ -1,0 +1,98 @@
+const path = require("path");
+
+const signatures = [
+  {
+    recipePath: recipe("javascript-class"),
+    conditions: {
+      tags: ["JavaScript", "Class"]
+    }
+  },
+  {
+    recipePath: recipe("javascript-constructor"),
+    conditions: {
+      tags: ["JavaScript", "Constructor"]
+    }
+  },
+  {
+    recipePath: recipe("javascript-error"),
+    conditions: {
+      tags: ["JavaScript", "Error"]
+    }
+  },
+  {
+    recipePath: recipe("javascript-method"),
+    conditions: {
+      tags: ["JavaScript", "Method"]
+    }
+  },
+  {
+    recipePath: recipe("javascript-property"),
+    conditions: {
+      tags: ["JavaScript", "Property"]
+    }
+  }
+];
+
+/**
+ * Add a path to a recipe file to `VFile.data.recipePath`.
+ *
+ * @returns A unified transformer.
+ */
+function annotateRecipePlugin() {
+  return function transformer(tree, file) {
+    const tagSet = new Set(file.data.tags);
+
+    for (const { recipePath, conditions } of signatures) {
+      if (hasAll(tagSet, conditions.tags)) {
+        file.data.recipePath = recipePath;
+        return;
+      }
+    }
+
+    const message = file.message(
+      "No recipe matched this page",
+      undefined,
+      "require-recipe"
+    );
+    message.fatal = true;
+  };
+}
+
+/**
+ * Convert a recipe name to a recipe file path.
+ *
+ * @param {String} name - the name of a recipe, without an extension
+ * @returns {String} An absolute path to a recipe file
+ */
+function recipe(name) {
+  return path.resolve(__dirname, "../../recipes", name + ".yaml");
+}
+
+/**
+ * Check if some tags contains some other tags.
+ *
+ * @param {Set<String>} actualTags
+ * @param {Set<String} expectedTags
+ * @returns {Boolean} `true` or `false`
+ */
+function hasAll(actualTags, expectedTags) {
+  return isSuperset(actualTags, expectedTags);
+}
+
+/**
+ * Check if a set is a subset of another set.
+ *
+ * @param {Set} set
+ * @param {Set} subset
+ * @returns {Boolean} `true` or `false`
+ */
+function isSuperset(set, subset) {
+  for (let elem of subset) {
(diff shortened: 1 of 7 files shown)


scripts/scraper-ng/plugins/annotate-recipe.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * Check if a set is a subset of another set.
+ *
+ * @param {Set} set
+ * @param {Set} subset
+ * @returns {Boolean} `true` or `false`
+ */
+function isSuperset(set, subset) {
```
wbamberg: I don't know why we have this and `hasAll()` which just calls this.

wbamberg:
Sorry I didn't process this, I'm distracted this week :(

> kuma-metadata: Is it a bad name or a terrible name?

I think it's fine.

> kuma-metadata: Should I just attach all the data from $json to the vfile?

I think what you have here is good.

> annotate-recipe warns about missing recipes. Should this be its own check in /scripts/scraper-ng/rules/ (e.g,. attempt to identify a recipe is one module, but requiring a recipe happens elsewhere)? If not, does this imply that rules and plugins should just be lumped together?

I tend to think it should be a separate check in /rules. So the plugin can set the recipe name if it can, and the rule should check whether a recipe was set. Otherwise, yes, it seems like the distinction between plugins (which seem kind of like preprocessors, or gatherers of data) and rules seems pretty blurry, and I do like the conceptual distinction.

But...

> annotate-recipe goes with the first recipe that matches, mostly because it was easy to write. Should recipe matches be required to be unique (i.e., a page should match one and only recipe?) or should we just concern ourselves with a best effort, by ranking recipe matchers by specificity?

I think each page should be unambiguously tagged as one and only one stumptown type. So a page tagged ["JavaScript", "Property", "Method"] is an error and it would be helpful for the linter to flag it. But that makes this separation between "plugins" and "rules" a bit trickier.

Maybe the "plugin" could just set the tags, then a rule could map that onto a recipe and generate an "invalid tags" error if a recipe could not be matched?

ddbeck:
> Also: very nit-picky, but "annotate-recipe" (and correspondingly annotateRecipePlugin) reads to me like: this plugin annotates the recipe.

Good point. I went with `identify-recipes`. More on the thinking behind this in a moment.

> Maybe the "plugin" could just set the tags, then a rule could map that onto a recipe and generate an "invalid tags" error if a recipe could not be matched?

I'd been mulling this over today and I think you're right that they should be separate. The guiding principle could be that plugins modify files or trees in some way, and rules only mark the passage of problems (they don't do anything to create or resolve those problems).

This means my plugin gets to assign a `recipePath`, but I won't actually check if the recipe makes sense later. The new code does this:

1. (unchanged) `kuma-metadata` sets the `vfile.data.tags` array.
2. `identify-recipes` adds `vfile.data.recipePath`. If only one recipe matches, this is just a path to a file. If more than one recipe matches, then this is an array of paths to recipes. In other words, I set a value with the intent of precipitating an error.
3. `file-require-recipe` checks that `vfile.data.recipePath` exists (i.e., it's not undefined—no recipes matched) and points to a file that exists. If it's an array or undefined, then I produce a pretty looking error.

I ran my new test with some filtering to hide unrelated errors and got something like this, about `/en-US/docs/Web/JavaScript`:

```
Summary
Missing recipe                        304 errors  file-require-recipe       
One and only one recipe must apply      6 errors  file-require-recipe-unique
<some network error>                    2 errors  null

3 notice types (312 notices, ✖ 312 errors) in 918 files
```

Errors about missing recipes look like this:

```
/en-US/docs/Web/JavaScript/New_in_JavaScript/1.8.5
  1:1  error  Missing recipe                                                    file-require-recipe
Tags: ["JavaScript 1.8.5","JavaScript","Versions","ECMAScript 5","Firefox 4"]
```

Errors about clashing recipes look like this:

```
/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Module/imports
  1:1  error  One and only one recipe must apply                                                          file-require-recipe-unique
Recipes: ["javascript-constructor","javascript-method"]
Tags: ["API","Constructor","Reference","Module","imports","JavaScript","Object","WebAssembly","Method"]
```

Seems pretty promising!

wbamberg:
> I set a value with the intent of precipitating an error.

So passive-aggressive, I love it.

> Seems pretty promising!

Yes, I think this is a great approach.

scripts/scraper-ng/plugins/identify-recipes.js
```diff
@@ -0,0 +1,86 @@
...
+  {
+    recipePath: recipe("javascript-constructor"),
+    conditions: {
+      tags: ["JavaScript", "Constructor"]
+    }
+  },
+  {
+    recipePath: recipe("javascript-error"),
+    conditions: {
+      tags: ["JavaScript", "Error", "Constructor"]
```
wbamberg: I should have seen this before, but why should these tags map to "javascript-error"? I would have expected `["JavaScript", "Errors"]` to catch pages like https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_formal_parameter.

ddbeck: Because I mistakenly thought the javascript-error recipe meant `Error`-like objects. I just pushed a limited fix.

That said, I don't want to get too stuck on the tag set matching before I've written matching for all the JavaScript recipes. It's not yet clear to me what will constitute an implementation problem with this script and what will constitute an upstream content issue. Just for starters, the set of tags `["JavaScript", "Errors"]` won't do the trick for pages under `/en-US/docs/Web/JavaScript/Reference/Errors`: https://gist.github.com/ddbeck/3253e48f2643cd23b6cd0aa978482fe2

wbamberg: > It's not yet clear to me what will constitute an implementation problem with this script and what will constitute an upstream content issue. 

Yes, that's a fair point. I just looked at one `Error` page and it had those tags. In fact `Error` seems like a better tag than `Errors` and it's quite possible that the pages use a mixture...

PR #289, opened 2020-01-17
Add scraper message filtering

ddbeck:
In the course of writing #288, I found it helpful to be able to filter linting messages by level and type. I'm opening this PR to see if it would make sense to make a formal part of the tooling.

(My alternative to this is a `--json` output option that you could pipe into other tools. I won't be heartbroken if one of my proposals is rejected.)

### DIFF
--- a/scripts/scraper-ng/index.js
+++ b/scripts/scraper-ng/index.js
@@ -18,6 +18,19 @@ const { argv } = yargs
   .example(`$0 ${examplePage}`, "scrape a page and all its subpages")
   .example(`$0 ${exampleShorthand}`, "omit the protocol and domain")
 
+  .describe("level", "Minimum message level")
+  .nargs("level", 1)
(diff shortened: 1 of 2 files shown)


ddbeck:
I discovered today this is broken in some amusing ways. It's not ready for review; I'll make a comment here when I've fixed it.

ddbeck:
OK, I've pushed two changes.

1. `--summary` threw an exception if no errors were reported (not related to the new options, but it's very noticeable with the new options—I can cherry pick that into a separate PR if needed). Fixed.
2. A bug caused _all_ messages to be ignored if you didn't use the `--ignore` option. Fixed.

ddbeck:
Based on #320, I think cleaning up the namespaces and rule IDs and then choosing rules to show or hide with a little less regex magic is the way to go. Closing this.

PR #296, opened 2020-01-30
fix scraper regex

wbamberg:
Make arguments optional in a macro call.


wbamberg:
I think we'll do this properly :)

Issue #294, opened 2020-01-29
javascript-namespace recipe


wbamberg:
We talked yesterday about having a separate page type for JS namespaces. This issue summarises our conversation as I understood it.

A namespace is distinguished from just a class because it is not instantiable. So for example, [`Math`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math), [`JSON`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON), [`Intl`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl), and [`WebAssembly`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly) would all be namespaces, not classes.

Some (but not all) namespaces also define constructors for other objects. For example, [`Intl.Collator()`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator) and [`WebAssembly.Memory()`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory).

In terms of recipes, it seems like [javascript-class.yaml](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-class.yaml) would be mostly unchanged except that I guess `data.class_constructor` would be made mandatory. Then there would be a new recipe "javascript-namespace.yaml" which could look something like:

```yaml
related_content: related_content/javascript.yaml
data:
  - title
  - short_title?
  - mdn_url
body:
  - prose.short_description
  - data.interactive_example?
  - prose.description?
  - data.classes?
  - data.static_properties?
  - data.static_methods?
  - data.instance_properties?
  - data.instance_methods?
  - prose.*
  - data.examples
  - data.specifications
  - data.browser_compatibility
  - prose.see_also
```

...so `data.classes?` is used to list any constructor defined in the namespace, like [`Intl.Collator()`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator). Or would `data.class_constructors?` be better, or is that confusingly too close to the [javascript-class](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-class.yaml) recipe?

@Elchi3 , @ddbeck , any opinions?

Elchi3:
[...]

PR #301, opened 2020-02-04
Add javascript-namespace recipe

wbamberg:
Per https://github.com/mdn/stumptown-content/issues/294 this adds a recipe for JS namespaces like `WebAssembly` and `Intl`.

I've chosen "constructor_properties" for the name of constructors in namespaces. It seems all right, and has the merit that the [`Intl` spec](https://tc39.es/ecma402/#sec-constructor-properties-of-the-intl-object) uses that terminology, as Florian notes in https://github.com/mdn/stumptown-content/issues/294#issuecomment-579663963.

I've also made constructors mandatory in the "javascript-class" recipe, because if there's no constructor, it's a namespace.

### DIFF
--- a/content/en-US/javascript/reference/classes/bigint/bigint.md
+++ b/content/en-US/javascript/reference/classes/bigint/bigint.md
@@ -2,6 +2,7 @@
 recipe: javascript-class
 title: 'BigInt'
 mdn_url: /en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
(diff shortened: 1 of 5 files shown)


wbamberg:
Making constructors mandatory for classes broke the build because BigInt was scraped before we separated out constructors. Commit 2 here adds a constructor, and support for build-json to know what to do with them.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #303, opened 2020-02-06
Update mapping from tags to recipes

wbamberg:
This PR updates the code used by the scraper to map from tags to recipes.

I've added mappings to "javascript-namespace", "javascript-language-feature", "landing-page", and "guide". I also changed the tag we want for error pages from "Errors" to "Error", to be consistent with the others.

A couple of points:
* instead of expecting the "Language feature" tag we could expect things like "Operator" or "Expression" and so on. This would probably result in us having to add fewer tags, but it seemed simpler to have a single new tag for this recipe, than to try to figure out exactly which tags should map to this recipe.
* "Landing page" and "Guide" are weird because they are not specific to JS. This means that eventually, we would need some way for the scraper to know that these are *JavaScript* guide pages, only so it can set the JS sidebar for them. For now at least I think it would be OK for the scraper to do this hardcoded based on the top-level path element.

### DIFF
--- a/scripts/scraper-ng/plugins/identify-recipes.js
+++ b/scripts/scraper-ng/plugins/identify-recipes.js
@@ -2,6 +2,12 @@ const path = require("path");
 
 const signatures = [
   {
+    recipePath: recipe("javascript-namespace"),
+    conditions: {
+      tags: ["JavaScript", "Namespace"]
+    }
+  },
+  {
     recipePath: recipe("javascript-class"),
     conditions: {
       tags: ["JavaScript", "Class"]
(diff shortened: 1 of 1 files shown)


ddbeck:
> instead of expecting the "Language feature" tag we could expect things like "Operator" or "Expression" and so on. This would probably result in us having to add fewer tags, but it seemed simpler to have a single new tag for this recipe, than to try to figure out exactly which tags should map to this recipe

Yeah, let's try this and see. If it seems to onerous to add the required tags, we can revisit this.

> This means that eventually, we would need some way for the scraper to know that these are _JavaScript_ guide pages, only so it can set the JS sidebar for them. For now at least I think it would be OK for the scraper to do this hardcoded based on the top-level path element.

Yeah, this seems like something that can happen as part of the scraping process. Guides and landing pages will need extra frontmatter compared to recipes that have it preset—seems like a special case that we can handle at the point of generating the frontmatter and not sooner.

PR #305, opened 2020-02-07
Use KumaScript's parser to parse KumaScript macros

ddbeck:
This PR uses KumaScript's parser to find macros. It's implemented as a parser that wraps `rehype-parser`. Here's how it works, roughly:

1. Before parsing begins, strip macro arguments from the doc source, using KumaScript's `parser.js`.
2. Parse the doc as a rehype document.
3. Restore the macro calls in the AST as special KumaScript macro nodes with the original macro arguments.

While I was here, I also borrowed the method that KumaScript uses to normalize macro names, to further increase consistency with KumaScript itself.

This solves a few problems at once, in order of severity:

* Some macro names are ignored. For example, if a macro contained a hyphen, then the existing regular expression didn't catch it. Since this implementation reuses KumaScript's parser, it necessarily recognizes the same macro names as KS itself.
* Macro arguments containing HTML are not recognized by the existing macro identification code. If a macro argument contains HTML, then rehype parsed the argument as if it were part of the document, rather than an arbitrary string. This implementation extracts macro arguments before rehype parses the page, then reinstates them after parsing.
* Macro call source locations were not precisely reported in the linting output. Because the KumaScript parser records the offset of macros within the document, this information can be passed into the rhype AST nodes for macros.

This approach has a couple drawbacks:

* The find-and-replace-and-find-and-replace-again process feels a little fragile. I've included some exception branches in the new parser to catch obvious inconsistencies (e.g., missing macros or length mismatches). But there may be other quirks lurking.
* KumaScript's `parser.js` is vendorized. `npm` doesn't have a good way to depend on an arbitrary JS module. We could probably turn `parser.js` into a proper package, but that wold need to be a separate PR against [mdn/kumascript](https://github.com/mdn/kumascript) and I'm not enthusiastic about potentially introducing bugs into KS itself. The good news is that the module rarely changes and I've tried to mitigate this problem by providing a package script to help update `parser.js`


ddbeck:
Oh, here's a few examples of what a difference this makes. Each is the number of macros that the two different branches finds:

| Macro             | `master` | this branch |
|-------------------|----------|-------------|
| `jsxref`          | 7279     | 7388        |
| `JSRef`           | 1        | 658         |
| `optional_inline` | 0        | 198         |

PR #316, opened 2020-02-20
ignore more

peterbe:
Every time something goes wrong with `npm run ...` I get this file that I have to ignore. 
I could put it into my personal `.git/info/exclude` or `~/.gitignore_global` but I think it benefits all users if this is ignored by default for everyone.


wbamberg (approved):
Looks good to me, thank you @peterbe !

PR #340, opened 2020-03-03
Fix bug a bug when printing a summary with no errors

ddbeck:
If you a lint a page that doesn't have any errors and request for the summary, then you get an exception and unhelpful output. This is sad and disappointing. This PR fixes the exception and prints a cheerful message instead.

### DIFF
--- a/scripts/scraper-ng/vfile-reporter-summary.js
+++ b/scripts/scraper-ng/vfile-reporter-summary.js
@@ -35,7 +35,7 @@ function summarize(files) {
     return [
       "",
       chalk.yellow(chalk.underline("Summary")),
-      "No issues found."
+      "No issues found. Hooray! 🎉"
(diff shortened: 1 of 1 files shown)


scripts/scraper-ng/vfile-reporter-summary.js
```diff
@@ -31,6 +31,14 @@ function reporter(files) {
 function summarize(files) {
   const count = countRules(files);
 
+  if (Object.entries(count).length === 0) {
+    return [
+      "",
+      chalk.yellow(chalk.underline("Summary")),
+      "No issues found. Hooray! 🎉"
```
wbamberg: Almost too cheerful.

ddbeck: Don't worry, you probably won't see it often

PR #345, opened 2020-03-06
Remove kumascript-macros plugin

ddbeck:
This has been superseded by `kumascript-rehype-parse.js`, added with #305's use of the KumaScript parser. I don't know why I forgot to remove it in the first place.


Elchi3 (approved):
This looks good to me, thank you! :+1: 

This made my stumble upon the scraper in `scraper/` where I guess this code has some roots from. Maybe we should remove `scraper/`? Or, I guess, we probably want to wait until `scraper-ng` actually creates markdown files for us.

ddbeck:
> This made my stumble upon the scraper in `scraper/` where I guess this code has some roots from. Maybe we should remove `scraper/`? Or, I guess, we probably want to wait until `scraper-ng` actually creates markdown files for us.

Yeah, I think `scraper-ng/` should move to `scraper/` eventually. I think @wbamberg may have used `scraper/` recently, so I don't think it's ready to go away until we're either _definitely_ not using it anymore or `scraper-ng/` supersedes it.

Issue #306, opened 2020-02-07
Support banners/headers


wbamberg:
See https://github.com/mdn/stumptown-renderer/issues/350.

I had thought that this would be doable with no changes in stumptown-content because the underlying info is in BCD. But it might be better to expose BCD's "status" in a more easily-assessed place in the JSON. My suggestion in that case would be to include it in the `body` object, as a new item like:

```json
{
    "type": "status",
    "value": [
        "non-standard", "deprecated"
    ]
}
```

...that is, a `type` of "status" and a `value` which is an array including any of  "non-standard", "deprecated", and "experimental". I think if none of these apply the whole item could be omitted.

Then this issue would be fixed by making updating build-json to pull that info out of the BCD, and adding this new item to the built JSON.

@Elchi3 , @ddbeck , what do you think?

wbamberg:
One thing is that BCD has "standard_track", which inverts the meaning from "non-standard". I think it would be nice to represent this in the JSON as "non-standard", though, since that's the case worth noting.

Also, it's a bit upsetting that "non-standard" duplicates information from the `specifications` property (see https://github.com/mdn/stumptown-content/issues/126#issuecomment-539448613). I don't know if we should try to address that.

Elchi3:
> Then this issue would be fixed by making updating build-json to pull that info out of the BCD, and adding this new item to the built JSON.

I think this would be great and I think short term this would get us to remove the macros nicely.


I also want to briefly give some more thoughts on banners or status properties. This stuff has been underspecified for a long time and so over in BCD, we've had discussions about how to best go about this. Some comments on how I see this:

- `deprecated` is mostly good as is. The issue that comes up is that sometimes it is a bit unclear who deprecated the feature and what it means. So, whether it is deprecation per spec or a browser deprecation and if it will ever be removed (usually it won't be removed for web compat reasons, but sometimes it will be [see app cache]).
- `obsolete` is a status term we've banned from MDN. It used to mean "deprecated and no longer implemented", but implementation status is compat data, so we've stopped using it. If you want to communicate "do not use!", use `deprecated`. I would remove all macros that call things "obsolete". The term is already banned in BCD (https://github.com/mdn/browser-compat-data/issues/265).
[...]

PR #354, opened 2020-03-11
Support 'generated.status' in build-json

wbamberg:
This PR addresses the content side of supporting banners/headers, as tracked by https://github.com/mdn/stumptown-content/issues/306.

It adds support for a new `status` ingredient. Its value is derived from the `specifications` and `browser-compatibility` properties, so is made a new sort of ingredient, `generated`: which means it appears in recipes but not in front matter: writers explicitly don't have to worry about it.

This is from @ddbeck 's suggestion here: https://github.com/mdn/stumptown-content/issues/306#issuecomment-596394564.

It's a bit inefficient because it queries the BCD twice for the data. But I think this is pretty quick and it didn't really seem worth caching it, just for this. But maybe it is? But anyway I don't want to cache *all* the BCD query lookups, as we do for [`related_content`](https://github.com/mdn/stumptown-content/blob/master/scripts/build-json/related-content.js#L31), because unlike `related_content`, BCD JSONs are not typically shared between different pages.

It'll add a new JSON object like:

```json
      "type": "status",
      "value": {
        "non_standard": true,
        "experimental": false,
        "deprecated": true
      }
    }
```

...placed at the start of the `body`:

```json
  "body": [
    {
      "type": "status",
      "value": {
        "non_standard": true,
        "experimental": false,
        "deprecated": true
      }
    },
    {
      "type": "prose",
      "value": {
        "title": "Short description",
        "id": "short_description",
        "content": "<p>...</p>"
      }
    },
```

### DIFF
--- a/scripts/build-json/resolve-status.js
+++ b/scripts/build-json/resolve-status.js
@@ -1,0 +1,12 @@
+function packageStatus(specifications, bcd) {
+  const bcdHasStatus = !!bcd.__compat && !!bcd.__compat.status;
+  return {
+    non_standard: specifications === "non-standard",
+    experimental: bcdHasStatus && bcd.__compat.status.experimental,
+    deprecated: bcdHasStatus && bcd.__compat.status.deprecated
+  };
+}
+
+module.exports = {
+  packageStatus
+};
--- a/scripts/build-json/resolve-status.js
+++ b/scripts/build-json/resolve-status.js
@@ -1,5 +1,6 @@
 function packageStatus(specifications, bcd) {
-  const hasStatus = bcd.__compat !== undefined && bcd.__compat.status !== undefined;
+  const hasStatus =
+    bcd.__compat !== undefined && bcd.__compat.status !== undefined;
   return {
     non_standard: specifications === "non-standard",
     experimental: hasStatus && bcd.__compat.status.experimental,
--- a/scripts/build-json/resolve-status.js
+++ b/scripts/build-json/resolve-status.js
@@ -1,11 +1,17 @@
 function packageStatus(specifications, bcd) {
   const hasStatus =
     bcd.__compat !== undefined && bcd.__compat.status !== undefined;
-  return {
-    non_standard: specifications === "non-standard",
-    experimental: hasStatus && bcd.__compat.status.experimental,
-    deprecated: hasStatus && bcd.__compat.status.deprecated
-  };
+  const status = [];
+  if (specifications === "non-standard") {
+    status.push("non_standard");
+  }
+  if (hasStatus && bcd.__compat.status.experimental) {
+    status.push("experimental");
+  }
+  if (hasStatus && bcd.__compat.status.deprecated) {
+    status.push("deprecated");
+  }
+  return status;
 }
 
 module.exports = {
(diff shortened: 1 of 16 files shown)


scripts/build-json/resolve-status.js
```diff
@@ -0,0 +1,12 @@
+function packageStatus(specifications, bcd) {
+  const bcdHasStatus = !!bcd.__compat && !!bcd.__compat.status;
```
ddbeck: `!!` is a bit cryptic. I'd prefer something more explicit here (`bcd.__compat !== undefined` or whatever we're getting at).

Also, the BCD schema requires `__compat` to have a `status` object. Seems like we should be able to trust BCD to not break Stumptown.

wbamberg: Thanks @ddbeck ! I'm glad BCD mandates `status`, and have correspondingly tried to simplify this bit of the code.

wbamberg: Of course this really wants optional chaining (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) but that's not in Node yet apparently.

scripts/build-json/resolve-status.js
```diff
@@ -0,0 +1,12 @@
+function packageStatus(specifications, bcd) {
+  return {
+    non_standard: specifications === "non-standard",
+    experimental:
+      bcd.__compat !== undefined && bcd.__compat.status.experimental,
+    deprecated: bcd.__compat !== undefined && bcd.__compat.status.deprecated
+  };
```
ddbeck: Sorry about this, I was wrong: https://github.com/mdn/browser-compat-data/pull/5816#issuecomment-598159285

Maybe something like this?

```suggestion
  const hasStatus = bcd.__compat !== undefined && bcd.__compat.status !== undefined;
  return {
    non_standard: specifications === "non-standard",
    experimental: hasStatus && bcd.__compat.status.experimental,
    deprecated: hasStatus && bcd.__compat.status.deprecated
  };
```

wbamberg: Sure. I don't expect webextensions will make it into Stumptown, but this is fine too.

wbamberg:
So we can visually check the rendered banners, I added a fake JS class "TestBanners", which borrows the compat data from [`Intl.ListFormat`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ListFormat), so it gets tagged "experimental". I've given it a constructor `TestBanners()` which also uses that compat data, but additionally is marked "non-standard", so we can see what multiple banners are like.

I've had to give them the `Array` URL, because one of the CI checks (which is helpfully not part of `npm test`) requires that `mdn_url` is a real MDN page.

wbamberg:
@ddbeck , in https://github.com/mdn/stumptown-renderer/pull/380, @peterbe asked for us to represent `status` as an array of constants rather than an object. This seems like a good idea, so   https://github.com/mdn/stumptown-content/pull/354/commits/24c96b789c3235391260f9fdd73b343b91bf056e tries to do that.

It also omits `status` if it's empty: in most cases it will be empty, and it does not seem worthwhile  shipping an empty array around, and it makes things more complicated for the renderer.

I know it seems/is pointless, but I'd appreciate a review if you have time, so we can close this issue. Thanks!

PR #355, opened 2020-03-11
Rename and revise linter rules and messages

### LINKED ISSUE #331: Make linter origins and rule IDs more useful and consistent
A follow up to #320 

[VFile messages](https://github.com/vfile/vfile-message) are used to organize warnings and errors in the linter. Messages support rule ids in namespaces, called [`origins`](https://github.com/vfile/vfile-message#vmessagereason-position-origin).

The current linter doesn't consistently use this feature. Confusingly, some rule IDs contain colons, which conflicts with the message constructors' colon-delimited namespace and rule ID processing. Additionally, better use of namespaces and IDs will help with filtering output.

To fix this, all messages should:

- Use a namespace that describes the scope of the message (e.g., `file-recipe`, `html-ingredient`, `html-macro`)
- Use a rule ID that identifies the specific error or warning (e.g., for `file-recipe`, `recipe-missing`, `one-recipe-only`, `recipe-file-existence`)
- Use slashes in rule IDs to indicate additional scopes, if needed. For example, `html-require-ingredient` rules can use a `recipe/ingredient/error` pattern.
- Use namespaces and rule IDs that are singular.

ddbeck:
This PR tries to clean up the linting messages by condensing some of the messages (to avoid hitting the max width of your terminal) and using a more consistent scheme for rule origins and IDs. This PR fixes #331.

In more detail, here's what I've tried to do with this PR:

* All rules get a namespace (`vfile-message` uses the word "source" for this, which may be a bit confusing, but I've tried to be consistent about this). Each namespace is named for the plugin that emits the message (e.g., `file-require-recipe` is the name of a plugin and, now, a namespace for rule IDs).
* Where practical, rule messages start with the same word or phrase within the same name space (e.g., all the recipe messages start with "Recipe").
* Messages were adjusted for length to try fit into narrower terminals.
* `html-warn-macros` has been renamed to `html-no-macros` to get the spirit of the rule and to remove the message level from the name of the module and namespace.

If you're using the JSON output, you can now filter against the name `ruleId` or `source`, to filter the rule or namespace, respectively:

```
"ruleId": "jsxref",
"source": "html-no-macros"
```

None of them should be `null` now.

### DIFF
--- a/scripts/scraper-ng/preset.js
+++ b/scripts/scraper-ng/preset.js
@@ -10,7 +10,7 @@ module.exports = {
     // For rules that don't need settings of any kind:
     //   [require('./rules/rule-name')]
     [require("./rules/file-require-recipe")],
-    [require("./rules/html-warn-macros"), allowedMacros],
+    [require("./rules/html-no-macros"), allowedMacros],
     [require("./rules/html-require-recipe-ingredients")]
   ]
 };
--- a/scripts/scraper-ng/rules/file-require-recipe.js
+++ b/scripts/scraper-ng/rules/file-require-recipe.js
@@ -12,7 +12,7 @@ function requireRecipe() {
     if (file.data.recipePath === undefined) {
       msg(
         file,
-        "Missing recipe",
+        "Recipe is missing",
         "recipe-missing",
         `Tags: ${JSON.stringify(file.data.tags)}`
       );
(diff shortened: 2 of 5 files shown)


wbamberg (approved):
Thanks @ddbeck . This does look much better.

PR #358, opened 2020-03-13
Modular ingredient handlers

### LINKED ISSUE #337: Organize linter ingredient handlers as modules
A follow up to #320

Complex ingredient handlers don't really work as inline functions in [`ingredient-handlers.js`](https://github.com/mdn/stumptown-content/blob/master/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers.js). The large ones (like `data.specifications` or `prose.short_description`) ought to be their own modules that are `require`d in `ingredient-handlers.js`.

A structure like this might work:

- `ingredient-handers/`
  - `data-specifications.js`
  - `index.js`
  - `prose-short-description.js`
  - `utils.js`

wbamberg:
This attempts to fix https://github.com/mdn/stumptown-content/issues/337.

Two main commits:
* https://github.com/mdn/stumptown-content/commit/e658eb2c51bb7f11de0ab02a6245260dd3a77687 factors out utility functions, which are expected to be used by >1 handler, into a separate module "utils.js". This also pushes `source` into the `context` object, because we need it in various places.
* https://github.com/mdn/stumptown-content/commit/ec2a79ddad18bc8ba05b61abe57418eb3355c0d7 factors out the three big ingredient handlers into their own modules, and also moves "ingredient-handlers.js" -> "ingredient-handlers/index.js" so everything can live in there.


scripts/scraper-ng/rules/html-require-recipe-ingredients/index.js
```diff
@@ -1,10 +1,11 @@
 const fs = require("fs");
 const path = require("path");
-
 const yaml = require("js-yaml");
```
ddbeck: Side note here: I've been grouping imports by standard library, external libraries, and then local modules. I don't know if we should have a style for this sort of thing, but that's what I was doing.

ddbeck (approved):
This looks good to me! Thank you, Will.

(I wasn't sure if maybe @Elchi3 also wanted a chance to review this, so I'm going to leave it unmerged for now. Feel free to merge this if it's blocking something else.)

scripts/scraper-ng/rules/html-require-recipe-ingredients/index.js
wbamberg: Oh I didn't realise that, but I'm happy for us to to this. I've put the blank line back :).

Elchi3 (approved):
Thank you, Will! :+1:

Issue #348, opened 2020-03-09
U - Write a linter ingredient handler for class and namespace members


wbamberg:
The linter should validate the following ingredients:

* data.class_properties
* data.static_methods
* data.static_properties
* data.instance_methods
* data.instance_properties

I think it should be practical to do these all together, since they are very similar.

## Acceptance criteria

The linter checks that 'javascript-class' and 'javascript-namespace' pages contain the correct structures for these ingredients, according to the specification agreed in https://github.com/mdn/sprints/issues/3075.

wbamberg:
I've been thinking a bit about this today and it raises a question. So far the linter only lints mandatory ingredients, so the ingredient must (1) exist and (2) be well-formed to pass, and there isn't really a separation between these questions.

But for optional ingredients: they don't have to exist, but if they do, they must be well-formed. So the linter has to ask, first, does this page try to include static methods (say) or an interactive example? If it doesn't, move on, but if it does, check it's valid.

A question here is: is it likely that a given ingredient will be mandatory in one recipe but optional in another? If the answer is no (which I think it is) we could push this problem into the ingredient handlers. So, say, the `data.static_methods` handler could have some heuristic to tell it whether this page seems to be trying to include this ingredient (e.g. does it contain the "Static methods" H2). If it doesn't, move on, otherwise validate (check the contents of the H2).

Then [`requireRecipeIngredientsPlugin`](https://github.com/mdn/stumptown-content/blob/master/scripts/scraper-ng/rules/html-require-recipe-ingredients/index.js#L11) could pass mandatory and optional ingredients into the handlers.

It might still be worth logging missing optional ingredients, to help us catch false negatives.

Thoughts @ddbeck , @Elchi3 ?

Elchi3:
> A question here is: is it likely that a given ingredient will be mandatory in one recipe but optional in another? If the answer is no (which I think it is) we could push this problem into the ingredient handlers. 

I also believe the answer is no. It looks like we have `prose.description?` and `prose.description` right now, but I think this is a bug. It should probably be optional everywhere.

> So, say, the `data.static_methods` handler could have some heuristic to tell it whether this page seems to be trying to include this ingredient (e.g. does it contain the "Static methods" H2). If it doesn't, move on, otherwise validate (check the contents of the H2).

I think this makes sense.

ddbeck:
Yeah, that sounds reasonable to me. If we wanted to support optional and required ingredients with the same handler, we could have the plugin pass an argument to the handler.

PR #361, opened 2020-03-17
Make JS description sections optional

Elchi3:
Makes "Description" sections optional throughout JS docs. When linting the whole JS tree:

Before:

```
Missing from javascript-method: prose.description                             143 errors     javascript-method/prose.description    
Missing from javascript-constructor: prose.description                        63 errors      javascript-constructor/prose.description   
Missing from javascript-property: prose.description                           22 errors      javascript-property/prose.description                                  
Missing from javascript-language-feature: prose.description                   19 errors      javascript-language-feature/prose.description                          
74 message types (9737 messages, ✖ 800 errors, ⚠ 8937 warnings) in 919 files
```

After:
```
70 message types (9490 messages, ✖ 553 errors, ⚠ 8937 warnings) in 919 files
```

I believe we should make description sections optional. Of course, this raises the question to authors when to have a "Description" section and when not. Currently, I believe, it is there when the method or any language feature needs more explanations that doesn't fit into the short description. Not all pages need these longer texts, though. Sometimes what's in "Description" could probably be put into (multiple) examples, but sometimes it is really just text and not code. So, my feeling is that the "Description" section is a bit like a guide page as a section in the reference. The author can freely make use of it to go deeper and explain concepts etc. More thoughts on this welcome.

Follows up on the comment I made here https://github.com/mdn/stumptown-content/issues/348#issuecomment-600036084

### DIFF
--- a/content/en-US/css/reference/properties/transform/transform.md
+++ b/content/en-US/css/reference/properties/transform/transform.md
@@ -16,7 +16,7 @@ examples:
 The **`transform`** [CSS](/en-US/docs/Web/CSS) property lets you rotate,
 scale, skew, or translate an element.
 
-## Overview
+## Description
 
 This property modifies the coordinate space
 of the CSS [visual formatting
--- a/content/en-US/html/HTML.md
+++ b/content/en-US/html/HTML.md
@@ -13,7 +13,7 @@ link_lists:
       - "/content/en-US/html/reference/elements"
 ---
 
-## Overview
+## Description
 
 **HTML** (HyperText Markup Language) is the most basic building block of the Web. It defines the meaning and structure of web content. Other technologies besides HTML are generally used to describe a web page's appearance/presentation ([CSS](/en-US/docs/Web/CSS)) or functionality/behavior ([JavaScript](/en-US/docs/Web/JavaScript)).
 
--- a/content/en-US/html/reference/elements/article/article.md
+++ b/content/en-US/html/reference/elements/article/article.md
@@ -32,7 +32,7 @@ in a document, page, application, or site, which is intended to be
 independently distributable or reusable (e.g., in syndication). Examples
 include: a forum post, a magazine or newspaper article, or a blog entry.
 
-## Overview
+## Description
 
 A given document can have multiple articles in it; for example, on a
 blog that shows the text of each article one after another as the reader
--- a/content/en-US/html/reference/elements/body/body.md
+++ b/content/en-US/html/reference/elements/body/body.md
@@ -21,7 +21,7 @@ recipe: html-element
 The **HTML `<body>` Element** represents the content of an HTML
 document.
 
-## Overview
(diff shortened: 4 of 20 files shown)


wbamberg:
I agree that lots of pages omitting a thing is a strong signal that it should be optional.

One question though is: in the recipes for [CSS properties](https://github.com/mdn/stumptown-content/blob/master/recipes/css-property.yaml) and for [HTML elements](https://github.com/mdn/stumptown-content/blob/master/recipes/html-element.yaml) we have mandatory "Overview" sections, that are supposed to follow the interactive example, if there is one. These sections are supposed to have special-case handling that omits the H2 (as for "short_description"). This is supposed to accommodate some text in between the interactive example and the named sections like "Attributes" (see for example https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address).

What's the difference between "Overview" and "Description" ? They seem to fulfill basically the same function.

I think my suggestion would be:
* let's use "Description" everywhere, and make it optional
* let's make it a normal section with an H2. Then only "short_description" is special-case, and I'm more comfortable with that.

Elchi3:
> What's the difference between "Overview" and "Description" ? They seem to fulfill basically the same function.

Thanks for thinking outside of JS docs! I agree "overview" fulfills basically the same function.

> I think my suggestion would be:
>  let's use "Description" everywhere, and make it optional
>  let's make it a normal section with an H2. Then only "short_description" is special-case, and I'm more comfortable with that.

I like this plan. To fully implement it, would there be follow-ups, or is any more work to be done in this PR?

wbamberg:
>  To fully implement it, would there be follow-ups, or is any more work to be done in this PR?

Why not, in this PR, update all the recipes that have `prose.overview` to have instead `prose.description?`? That gives us a consistent base.

Elchi3:
> Why not, in this PR, update all the recipes that have prose.overview to have instead prose.description?? That gives us a consistent base.

Done that now, for some reason I though this would be more complex but looks like it isn't. I hope it is reviewable.

Some unrelated stuff seem to have made it into the commit. Wasn't there when I reviewed the diff. I guess it is this magic?

> husky > pre-commit (node v12.16.1)
> 🔍  Finding changed files since git revision 75204b2.
> 🎯  Found 8 changed files.
> ✍️  Fixing up project-docs/html-element.md.
> ✍️  Fixing up project-docs/stumptown-status.md.
> ✍️  Fixing up project-docs/stumptown-writers-guide.md.
> ✅  Everything is awesome!

ddbeck:
> Some unrelated stuff seem to have made it into the commit. Wasn't there when I reviewed the diff. I guess it is this magic?

This is because the husky pre-commit command applies prettier in ways that the package scripts prettier commands don't—in this case, it's running on all files, including Markdown. I'm not sure when that got added or why, but that's what's doing it.

Issue #375, opened 2020-03-26
Inconsistency between recipe and document script


jpmedley:
Linting the stumptown repo produces warnings for section that are clearly in the documents being tested. In the example below, set.md actually contains sections named "description", "see_also", "short_description", and "syntax". Yet the linter says they are missing.

![image](https://user-images.githubusercontent.com/8204171/77660001-028e0380-6f36-11ea-8138-d1178b1f6103.png)

I stepped through the code and found the place where found sections are checked against the recipe. For the purpose of linting, sections in tested documents are represented according to heading level. For example, markdown that says `## Description` is represented as `h2.description` in code. Unfortunately, the linter is looking for `description`. The [exact point where](https://github.com/mdn/stumptown-content/blob/master/scripts/linter/plugins/missing-sections.js#L40) the comparison happens and where the warning is written is shown below.

I'm happy to open a PR to fix this. But I need to know what to do.
* Should I edit the recipes so that items like `prose.description` are changed to `prose.h2.description`? (I've already determined that this fixes the problem.)
* Or, should I change the code so that linting is done on `description` instead of `h2.description`?

![image](https://user-images.githubusercontent.com/8204171/77660253-51d43400-6f36-11ea-8ca0-e3ce00a640f8.png)

wbamberg:
I believe you should do this:

> change the code so that linting is done on description instead of h2.description?

I think the recipe is right, the linter is wrong. But @ddbeck wrote this code, so it's worth asking him too.

ddbeck:
I agree that it's a bug with the linter code, not the recipe. 👍

PR #391, opened 2020-04-07
Add linter requirements for JavaScript documentation

wbamberg:
This document describes what we want to lint for the JavaScript reference documentation.

### DIFF
--- a/project-docs/javascript-linter-spec.md
+++ b/project-docs/javascript-linter-spec.md
@@ -134,7 +134,7 @@ To satisfy this ingredient a page must have a section demarcated by `H2#Examples
 
 Each of these examples may include "code sections". These are marked up as follows:
 
-- HTML code section: `<pre class="brush: html language-html"><code class=" language-html">`
+- HTML code section: `<pre class="brush: html language-html"><code class="language-html">`
 - CSS code section: `<pre class="brush: css language-css"><code class=" language-css">`
 - JavaScript code section: `<pre class="brush: js language-js"><code class=" language-js">`
 
--- a/project-docs/javascript-linter-spec.md
+++ b/project-docs/javascript-linter-spec.md
@@ -156,7 +156,7 @@ optional JavaScript section
 
 #### data.specifications
 
-To satisfy this ingredient a page must have a section demarcated by `H2#Specifications`. It must contain either a call to the `{{SpecName}}` macro, or a text node containing the text "Not part of any standard".
+To satisfy this ingredient a page must have a section demarcated by `H2#Specifications`. It must contain either a call to the `{{SpecName}}` macro, or a text node containing the text "Not part of any standard.".
 
 #### data.browser_compatibility
 
--- a/project-docs/javascript-linter-spec.md
+++ b/project-docs/javascript-linter-spec.md
@@ -20,7 +20,7 @@ _the items in the page must correspond to the ingredients listed in the recipe, 
 
 A recipe can mark an ingredient as optional by placing a question mark after it, like `prose.description?`. This means that the page does not have to contain a representation of the ingredient, but if it does, the representation must be properly formed according to the ingredient's specification, and the representation must appear in the place specified by the recipe.
 
-Technically, whether an ingredient is optional or not is a separate attribute of the ingredient. This means that a single ingredient may be marked as optional in one recipe, and mandatory in a different recipe. In practice, in the JavaScript documentation, this does not occur: there are currently no ingredients that are optional in some recipes and mandatory in others.
+Technically, whether an ingredient is optional or not is a separate attribute of the ingredient. This means that a single ingredient may be marked as optional in one recipe and mandatory in a different recipe. In practice, in the JavaScript documentation, this does not occur: there are currently no ingredients that are optional in some recipes and mandatory in others.
 
 ### Page sections
 
@@ -95,7 +95,7 @@ This is an optional prose ingredient. To satisfy the ingredient, if the page con
 
 #### prose.\*
 
-This ingredient enables writers to include extra sections, but ensuring that they always appear in the same place in the page. To satisfy this ingredient: if the page contains any `<h2>` headings that are not expected by the recipe, then they must appear at the point in the page that the `prose.*` ingredient is listed in the recipe.
+This ingredient enables writers to include extra sections that are not explicitly listed in the recipe, while ensuring that the extra sections always appear in the same place in the page. To satisfy this ingredient: if the page contains any `<h2>` headings that are not expected by the recipe, then they must appear at the point in the page that the `prose.*` ingredient is listed in the recipe.
 
 ### Data ingredients
 
--- a/project-docs/javascript-linter-spec.md
+++ b/project-docs/javascript-linter-spec.md
@@ -14,6 +14,12 @@ So at a high level, compliance to a recipe means:
 
 _the items in the page must correspond to the ingredients listed in the recipe, in the correct order, and each item must satisfy the particular requirements for its ingredient_.
 
+### Optional ingredients
+
+A recipe can mark an ingredient as optional by placing a question mark after it, like `prose.description?`. This means that the page does not have to contain a representation of the ingredient, but if it does, the representation must be properly formed according to the ingredient's specification, and the representation must appear in the place specified by the recipe.
+
+Technically, whether an ingredient is optional or not is a separate attribute of the ingredient. This means that a single ingredient may be marked as optional in one recipe, and mandatory in a different recipe. In practice, in the JavaScript documentation, this does not occur: there are currently no ingredients that are optional in some recipes and mandatory in others.
+
 ## JavaScript recipes
 
 We've defined the following recipes for the JS docs:
--- a/project-docs/javascript-linter-spec.md
+++ b/project-docs/javascript-linter-spec.md
@@ -96,7 +96,7 @@ This is an optional data ingredient. To satisfy the ingredient, if the page cont
 To satisfy this ingredient, a page must meet the following requirements:
 
 - A section demarcated by `H2#Constructor` that contains only a `<dl>` element.
-- The `<dl>` must contain a single `<dt><dd>` pair.
+- The `<dl>` must contain a single `<dt>` followed by a single `<dd>`.
 - The `<dt>` must contain either:
   - only a single `<code>` element, that contains only a single `<a>` element
   - only a call to the `{{jsxref}}` macro.
--- a/project-docs/javascript-linter-spec.md
+++ b/project-docs/javascript-linter-spec.md
@@ -114,6 +114,7 @@ To satisfy this ingredient, a page must meet the following requirements:
 - The `<dt>` must contain either:
   - only a single `<code>` element, that contains only a single `<a>` element
   - only a call to the `{{jsxref}}` macro.
+- The `<dd>` must contain `Creates a new <code>NameOfTheObject</code> object.` where `NameOfTheObject` is replaced with the actual name.
 
 #### data.constructor_properties?
 
--- a/project-docs/javascript-linter-spec.md
+++ b/project-docs/javascript-linter-spec.md
@@ -1,0 +1,163 @@
+# JavaScript reference docs structure specification
+
+This document describes the structure we want to require for JavaScript reference documentation.
+
+## Recipes and ingredients overview
+
+It is expected that pages documenting different sorts of thing will need to contain different elements. For example, the page for the JavaScript `BigInt` class will want to contain lists of links to its members, while the page for the JavaScript `BigInt.asUintN()` static method will want to contain a section describing the method syntax.
+
+To reflect this, a "recipe" is a short YAML file that describes at a high level the elements that a particular type of page is expected to have. A recipe provides a list of elements, that are expected to appear in the page in the order given. So for each distinct type of page, we define a single recipe.
+
+Each element is called an "ingredient". Ingredients are often shared across recipes. So for example we have an ingredient called `data.browser_compatibility`, which most recipes will contain. Each ingredient may have particular requirements about how it is represented in the page: for example, to satisfy the `data.browser_compatibility` ingredient a page must include the `{{Compat}}` KumaScript macro.
+
+So at a high level, compliance to a recipe means:
+
+_the items in the page must correspond to the ingredients listed in the recipe, in the correct order, and each item must satisfy the particular requirements for its ingredient_.
+
+## JavaScript recipes
+
+We've defined the following recipes for the JS docs:
+
+- [javascript-class](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-class.yaml)
+- [javascript-namespace](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-namespace.yaml)
+- [javascript-constructor](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-constructor.yaml)
+- [javascript-method](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-method.yaml)
+- [javascript-property](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-property.yaml)
+- [javascript-error](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-error.yaml)
+- [javascript-language-feature](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-language-feature.yaml)
+
+## JavaScript recipe ingredients
+
+In total these recipes use the following ingredients:
+
+prose ingredients:
+
+- prose.short_description
+- prose.description?
+- prose.see_also
+- prose.syntax
+- prose.message
+- prose.error_type
+- prose.what_went_wrong
+- prose.\*
+
+data ingredients:
+
+- data.interactive_example?
+- data.class_constructor
+- data.constructor_properties?
+- data.static_properties?
+- data.static_methods?
+- data.instance_properties?
+- data.instance_methods?
+- data.examples
+- data.specifications
+- data.browser_compatibility
+
+All ingredients except two are demarcated in the page using an `<h2>` heading. The exceptions are `prose.short_description` and `data.interactive_example`.
+
+### Prose ingredients
+
+With three exceptions, all prose ingredients impose the same requirement on the page: it must have an `<h2>` section whose ID attribute maps to the named section. The mapping is simply that `prose.` is removed, and the first letter is capitalized:
+
+```
+prose.syntax      ->    Syntax
+prose.see_also    ->    See_also
+```
+
+The exceptions are `prose.short_description`, `prose.description?`, and `prose.*`.
+
+#### prose.short_description
+
+This ingredient isn't demarcated by an `<h2>` at all. To satisfy the ingredient, a page must:
+
+- contain a `<p>` element
+- that is not a note or a warning
+- before the interactive example or the first `<h2>`, whichever comes first
+
+#### prose.description?
+
+This is an optional prose ingredient. To satisfy the ingredient, if the page contains an `h2#Description`, then it must appear in the place specified by the recipe.
+
+#### prose.\*
+
+This ingredient enables writers to include extra sections, but ensuring that they always appear in the same place in the page. To satisfy this ingredient: if the page contains any `<h2>` headings that are not expected by the recipe, then they must appear at the point in the page that the `prose.*` ingredient is listed in the recipe.
+
+### Data ingredients
+
+Data ingredients typically impose more detailed requirements on pages than prose ingredients.
+
+#### data.interactive_example?
+
+This is an optional data ingredient. To satisfy the ingredient, if the page contains a call to the `{{EmbedInteractiveExample}}` macro, then the call must appear in the place specified by the recipe.
+
+#### data.class_constructor
+
+To satisfy this ingredient, a page must meet the following requirements:
+
+- A section demarcated by `H2#Constructor` that contains only a `<dl>` element.
+- The `<dl>` must contain a single `<dt><dd>` pair.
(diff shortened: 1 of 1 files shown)


project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+This ingredient isn't demarcated by an `<h2>` at all. To satisfy the ingredient, a page must:
+
+- contain a `<p>` element
+- that is not a note or a warning
+- before the interactive example or the first `<h2>`, whichever comes first
+
+#### prose.description?
+
+This is an optional prose ingredient. To satisfy the ingredient, if the page contains an `h2#Description`, then it must appear in the place specified by the recipe.
```
Elchi3, line 80: Should we explicitly say that the question mark makes ingredients optional? Maybe this duplicates what we have in the writing guide, so maybe we should just link to it somewhere, so that people can learn about recipes first, if needed? https://github.com/mdn/stumptown-content/blob/master/project-docs/stumptown-writers-guide.md#recipe-files

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
+# JavaScript reference docs structure specification
+
+This document describes the structure we want to require for JavaScript reference documentation.
+
+## Recipes and ingredients overview
+
+It is expected that pages documenting different sorts of thing will need to contain different elements. For example, the page for the JavaScript `BigInt` class will want to contain lists of links to its members, while the page for the JavaScript `BigInt.asUintN()` static method will want to contain a section describing the method syntax.
```
Elchi3, line 7: I wasn't sure if it is "sorts of thing**s**"?
This then made me read up https://english.stackexchange.com/questions/68599/this-kind-of-things-vs-these-kinds-of-thing and now I wonder if we can instead say "It is expected that pages documenting different things will need to contain different elements." to avoid this altogether.

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### data.class_constructor
+
+To satisfy this ingredient, a page must meet the following requirements:
+
+- A section demarcated by `H2#Constructor` that contains only a `<dl>` element.
+- The `<dl>` must contain a single `<dt><dd>` pair.
+- The `<dt>` must contain either:
+  - only a single `<code>` element, that contains only a single `<a>` element
+  - only a call to the `{{jsxref}}` macro.
```
Elchi3, line 102: If we are getting this specific, I also tried to always have this format in the `<dd>` element: 

`"Creates a new <code>NameOfTheObject</code> object."`

Maybe I'm going too far now, though :)

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+This is an optional ingredient. It's just the same as `data.constructor_properties?`, except the section is demarcated by `H2#Instance_methods`.
+
+#### data.examples
+
+To satisfy this ingredient a page must have a section demarcated by `H2#Examples`. It must contain one or more sections demarcated by an H3. Each of these sections represents a single example.
+
+Each of these examples may include "code sections". These are marked up as follows:
+
+- HTML code section: `<pre class="brush: html language-html"><code class=" language-html">`
```
Elchi3, line 137: Are these spaces here and for CSS and JS on purpose?

```suggestion
- HTML code section: `<pre class="brush: html language-html"><code class="language-html">`
```

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+```
+unstructured prose
+optional HTML section
+optional CSS section
+optional JavaScript section
+```
+
+#### data.specifications
+
+To satisfy this ingredient a page must have a section demarcated by `H2#Specifications`. It must contain either a call to the `{{SpecName}}` macro, or a text node containing the text "Not part of any standard".
```
Elchi3, line 159: I see this is without the full stop in the linter code, too, but should it be required?

```suggestion
To satisfy this ingredient a page must have a section demarcated by `H2#Specifications`. It must contain either a call to the `{{SpecName}}` macro, or a text node containing the text "Not part of any standard.".
```

Elchi3 (approved):
Thanks Will! This reads great to me! Happy to merge as is. The comments I made are _very_ minor.

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+This ingredient enables writers to include extra sections, but ensuring that they always appear in the same place in the page. To satisfy this ingredient: if the page contains any `<h2>` headings that are not expected by the recipe, then they must appear at the point in the page that the `prose.*` ingredient is listed in the recipe.
+
+### Data ingredients
+
+Data ingredients typically impose more detailed requirements on pages than prose ingredients.
+
+#### data.interactive_example?
+
+This is an optional data ingredient. To satisfy the ingredient, if the page contains a call to the `{{EmbedInteractiveExample}}` macro, then the call must appear in the place specified by the recipe.
```
ddbeck, line 92: This definition is short and extremely complicated. For most ingredients, I can take a slice of a page and tell you if it contains the ingredient (without a recipe, even). But I could never write a test for this ingredient alone.

I'm kind of uncomfortable with the idea that an ingredient is undefined until it's in a recipe. Do we lose something if we say that `data.interactive_example?` is an `{{EmbedInteractiveExample}}` macro call in a block by itself?

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+- that is not a note or a warning
+- before the interactive example or the first `<h2>`, whichever comes first
+
+#### prose.description?
+
+This is an optional prose ingredient. To satisfy the ingredient, if the page contains an `h2#Description`, then it must appear in the place specified by the recipe.
+
+#### prose.\*
+
+This ingredient enables writers to include extra sections, but ensuring that they always appear in the same place in the page. To satisfy this ingredient: if the page contains any `<h2>` headings that are not expected by the recipe, then they must appear at the point in the page that the `prose.*` ingredient is listed in the recipe.
```
ddbeck, line 84: > any `<h2>` headings that are not expected by the recipe

Interesting! I had previously understood `prose.*` to mean: any `<h2>`-demarcated sections that are not otherwise _specified_. I think your definition may might have implications for _optional_ data ingredients—though I guess that depends on the earlier optionality question. This is complicated and I'm not sure if it matters yet, but I thought I should write it down. 😆

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+#### data.interactive_example?
+
+This is an optional data ingredient. To satisfy the ingredient, if the page contains a call to the `{{EmbedInteractiveExample}}` macro, then the call must appear in the place specified by the recipe.
+
+#### data.class_constructor
+
+To satisfy this ingredient, a page must meet the following requirements:
+
+- A section demarcated by `H2#Constructor` that contains only a `<dl>` element.
+- The `<dl>` must contain a single `<dt><dd>` pair.
```
ddbeck, line 99: I understood this to mean:

```suggestion
- The `<dl>` must contain a single `<dt>` followed by a single `<dd>`.
```

Is that right? Is that more clear or is that unhelpful?

ddbeck (commented):
Thank you for writing this, Will! It really got me thinking about what we mean we talk about ingredients and recipes. I tried to think as if I were implementing the spec as I read this. I've raised a few issues here. I admit some are quite large, so I'm happy to open issues for anything too big to address quickly.

project-docs/javascript-linter-spec.md
wbamberg: This is what the markup currently looks like, I just copy/pasted it. But you're right, there's no reason to have it here that I can see.

project-docs/javascript-linter-spec.md
wbamberg: I don't follow. Can you elaborate on the difference between what you thought before and what you think this is saying? It's just as possible that I'm expressing it wrong here (and very likely that this isn't clearly enough expressed, since we have doubt about what it means).

ddbeck: OK, now that I've written this out, I'm less worried about this thing. But here's what I was thinking:

I thought `prose.*` would be any section that would not otherwise appear as part of a recognized ingredient. For example, suppose you had a page with a "Specifications" section with the proper macro call, but that page's recipe did not include `data.specifications`. It seems to me that we could reasonably warn about that because the page doesn't call for that ingredient.

But `prose.*` complicates matters. Under your definition, the presence of `prose.*` in the recipe might allow for content that matches the `data.specifications` ingredient, even if that ingredient is not part of the recipe. In other words, there's no way to _forbid_ `data.specifications` content from appearing in a page, if it's in a range of sections consumed by `prose.*`.

In my definition, `prose.*` could _never_ consume a "Specifications" H2; a "Specifications" H2 could _only_ be used to mark the start of content that adheres to `data.specifications`. Now that I've spelled this out, it's probably a bad idea because you'd have to contort yourself (e.g., writing "Notes on specifications") to avoid clashing with section headings that are already "taken" by an ingredient.

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### prose.short_description
+
+This ingredient isn't demarcated by an `<h2>` at all. To satisfy the ingredient, a page must:
+
+- contain a `<p>` element
+- that is not a note or a warning
+- before the interactive example or the first `<h2>`, whichever comes first
+
+#### prose.description?
```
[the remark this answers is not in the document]
wbamberg: Good question! I agree, the "?" is an annotation, the ingredient is really `prose.description`. We could definitely imagine the same ingredient being mandatory in some recipes and optional in others.

However, at the moment, in practice, we don't have this situation, so it seemed a reasonable short-cut not to implement optional-ness separately (as discussed in https://github.com/mdn/stumptown-content/issues/348). Similarly it seemed that to describe them as a single unit in this specification was more direct and helpful for implementors.

Perhaps it isn't helpful to do this though? Or perhaps if the spec explained the above reasoning (along with Florian's comment at https://github.com/mdn/stumptown-content/pull/391#discussion_r405520364) then that would be OK?

project-docs/javascript-linter-spec.md
wbamberg: The difficulty is that the ingredient is optional. So while "an `{{EmbedInteractiveExample}}` macro call in a block by itself" satisfies `data.interactive_example?`, so does "".

Sometimes (for example, for `data.static_methods`) we manage to distinguish three cases:

1. the page omits the ingredient (OK for optional ingredients)
2. the page tries to include the ingredient but specifies it wrong (not OK)
3. the page includes the ingredient and gets it right (OK for optional and mandatory ingredients)

Even there this is a hard line to maintain. If the page contains `H2#Static_Methods` is this case 1 or case 2? What about `H2#Static_functions`. And so on. Realistically, there's just going to be some stuff we get wrong.

But with `data.interactive_example?` it's hard to see any space to accommodate case 2.

> For most ingredients, I can take a slice of a page and tell you if it contains the ingredient (without a recipe, even). But I could never write a test for this ingredient alone.

This gets into the question above of whether the "?" is part of the ingredient or not. If it's not, then you can easily write a test for whether this section "contains the ingredient" - does it contain `{{EmbedInteractiveExample}}`? But when the ingredient is optional that's not a useful test. So it seems that the test for a recipe containing `data.interactive_example?` just collapses into the requirement for the ingredient (i.e. the macro call) to appear in the given place.

project-docs/javascript-linter-spec.md
wbamberg: In practice I can't think of a place we're likely to want a page to use a heading that's not in its recipe but is in another recipe. But it doesn't seem like a thing we want to exclude.

I suppose one factor here is: the more heterogeneous are the different doc types that we expect this system to govern, then the less sense it seems to make for one recipe to be able to prevent another recipe from using a particular H2.

project-docs/javascript-linter-spec.md
wbamberg: I added a section on optional ingredients: https://github.com/mdn/stumptown-content/pull/391/commits/79b22c7931d8120fe1c3c1bc810517d6b623c4ed. Given all the ongoing weirdness around stumptown and Markdown, I want this doc to be a single self-contained item describing linting to recipes, so I don't really want to link out to general stumptown docs. (note that it does not mention Markdown or stumptown at all).

project-docs/javascript-linter-spec.md
wbamberg: Well... "different sorts of thing" is supposed to mean, say, methods or properties or classes. "different things" could be, say, `Array`, `BigInt`, `Date` - which are the same sort of thing.

I've tried to clarify things in this commit: https://github.com/mdn/stumptown-content/pull/391/commits/1e3a35e224bbf7e71c6318085cc1ec8b72d5dbca - but might have made it worse!

BTW whenever I read threads like that stackexchange thing (1) I'm so sorry for people who have to learn English and (2) I realise I don't understand any of it.

me-fail-english

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
+# JavaScript reference docs structure specification
+
+This document describes the structure we want to require for JavaScript reference documentation.
+
+## Recipes and ingredients overview
+
+It is expected that pages documenting different sorts of thing will need to contain different elements. For example, the page for the JavaScript `BigInt` class will want to contain lists of links to its members, while the page for the JavaScript `BigInt.asUintN()` static method will want to contain a section describing the method syntax.
+
+To reflect this, a "recipe" is a short YAML file that describes at a high level the elements that a particular type of page is expected to have. A recipe provides a list of elements, that are expected to appear in the page in the order given. So for each distinct type of page, we define a single recipe.
```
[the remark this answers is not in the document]
wbamberg: >  I'm not asking you to do this unless you're a little excited about it.

How much is a little ? :)

I'm happy to have a go at this but think it will be less confusing if we get through the substantive comments first.

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,163 @@
...
+- data.instance_methods?
+- data.examples
+- data.specifications
+- data.browser_compatibility
+
+All ingredients except two are demarcated in the page using an `<h2>` heading. The exceptions are `prose.short_description` and `data.interactive_example`.
+
+### Prose ingredients
+
+With three exceptions, all prose ingredients impose the same requirement on the page: it must have an `<h2>` section whose ID attribute maps to the named section. The mapping is simply that `prose.` is removed, and the first letter is capitalized:
```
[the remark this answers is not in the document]
wbamberg: I have had a go at this in https://github.com/mdn/stumptown-content/pull/391/commits/db631f70db0c2f7cab62f13005d15ceeb19265af.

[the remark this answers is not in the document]
ddbeck: This is good. Thank you!

project-docs/javascript-linter-spec.md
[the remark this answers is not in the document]
ddbeck: I like the way you've handled this with the new section. I think you're right that it's significant that we're using ingredients as either optional or not across all the JS recipes. The changes you've made on this work for me.

project-docs/javascript-linter-spec.md
ddbeck: Hmm. Maybe the right thing to do here is experiment with the more restrictive option. That would show us whether (and how) it breaks for the JS docs (I don't think this is likely, because they're so alike). When we expand to other areas, we may find that `prose.*` needs to be redefined, or that we need to introduce another, more permissive ingredient type.

project-docs/javascript-linter-spec.md
ddbeck: Can we define the interactive example by its position in the page without reference to other ingredients? Since we only use interactive examples after short descriptions, we could bake that assumption into the ingredient. Something like this:

> To satisfy the ingredient, a page must have a a `<div>` containing an `{{EmbedInteractiveExample}}` macro call. The `<div>` must be immediately preceded by paragraph that is neither a note nor a warning. The `<div>` must precede any `<h2>`.

project-docs/javascript-linter-spec.md
Elchi3: Thanks Will! Agree it's a good idea to be self-contained. Also great you now talk about the fact that JS ingredients aren't optional in some JS recipes and mandatory in others. I thought this was an interesting discovery during our work in the last sprint.

project-docs/javascript-linter-spec.md
Elchi3: You made it better :) Thanks for talking about "page types". I often think of page types when I think of recipes, so thanks for mentioning this here.

> BTW whenever I read threads like that stackexchange thing (1) I'm so sorry for people who have to learn English and (2) I realise I don't understand any of it.

Thanks for your sympathy! :smile: This makes me feel less stupid with English.

project-docs/javascript-linter-spec.md
wbamberg: I have added this: https://github.com/mdn/stumptown-content/pull/391/commits/725dba8d4ad88ec9f17be4c3c93052eda5d00a98 . I'm a bit worried it will be hard to test, especially getting `NameOfTheObject`, but let's revise it later if it looks too hard.

project-docs/javascript-linter-spec.md
wbamberg: This spec exists to serve us, so if that's useful I'm happy to add it. But although your suggestion does satisfy the ingredient, so does "the page does not contain a call to {{EmbedInteractiveExample}}". Would you be OK with:

> To satisfy the ingredient, either:
> * the page must not contain a `{{EmbedInteractiveExample}}` macro call, or
> * the page must have a `<div>` containing an `{{EmbedInteractiveExample}}` macro call. The `<div>` must be immediately preceded by paragraph that is neither a note nor a warning. The `<div>` must precede any `<h2>`.

project-docs/javascript-linter-spec.md
wbamberg: I don't think it's likely to break for the JS docs either, so this has no real practical effect. But it just seems wrong for recipe A to have any effect on what recipe B can include.

Another way to think of it: at the moment there are two categories of prose H2 headings:

* ones that are explicitly defined in its recipe: it must have these (or may have them if the ingredient is optional), and they must be in the defined place
* ones that are not explicitly defined in its recipe, which the page is allowed if `prose.*` is present

This adds a third category:
* ones it's just not allowed to use, regardless of `prose.*`, because an unrelated recipe is using them over in the SVG docs or something

I don't see the rationale for this category. It makes the rules more complex, and means authors have to know about all the recipes to write correct docs, rather than just their own (it makes recipes correctness dependent on other recipes).

But, like I say, I think this has no practical effect on the JS docs either way, so I will make this change if you're still not convinced.

project-docs/javascript-linter-spec.md
ddbeck: Yeah, this would work for me. Making, for example, an ingredient handler for this seems way easier than the one that has to know about the other ingredients to resolve. 👍

project-docs/javascript-linter-spec.md
ddbeck: Yeah, it's fine to leave this as is. I don't think this obligates authors to know about all the recipes, but it would obligate them to know about all the possible ingredients. That's still kinda annoying.

But it's also not essential: just because `prose.*` is specified permissively, doesn't mean we can't have additional linting that finds possibly inadvisable content. Here's a (somewhat contrived) example: imagine you've got a JS namespace page that accidentally contains a class constructor section. Depending on where that class constructor section appears in the page, it can be a valid page, even though the content is garbage no matter its position. But there's nothing stopping us from warning for unexpected ingredients, even if the spec says it's ultimately valid.

project-docs/javascript-linter-spec.md
wbamberg: OK -> https://github.com/mdn/stumptown-content/pull/391/commits/8222f7a2fa54b6eff281ad73ae2245fc048680e6

project-docs/javascript-linter-spec.md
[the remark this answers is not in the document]
wbamberg: -> https://github.com/mdn/stumptown-content/pull/391/commits/27178729dd1d32f302a664348cf2e5a8e8de756e

I made some minor changes here but didn't think it was worth going for some full-scale rewrite, especially as we have no actual template for requirements docs, so I would really be guessing about the exact language to use.

ddbeck (approved):
Looking good. Thanks for all the revisions on this, Will.

wbamberg:
Thanks for being so engaged with this!

jpmedley (commented):
Apparently, I wrote but neglected to post this comment.

project-docs/javascript-linter-spec.md
```diff
@@ -0,0 +1,178 @@
...
+- data.constructor_properties?
+- data.static_properties?
+- data.static_methods?
+- data.instance_properties?
+- data.instance_methods?
+- data.examples
+- data.specifications
+- data.browser_compatibility
+
+All ingredients except two are demarcated in the page using an `<h2>` heading. The exceptions are `prose.short_description` and `data.interactive_example`.
```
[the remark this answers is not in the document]
wbamberg: Thanks for the comment Joe. I somehow missed your bug report although you did tell me about it. I've commented there.

I think you know this too, but to be clear: the linter you're referring to in https://github.com/mdn/stumptown-content/issues/375 is the tool that lints Markdown in GitHub. This linter is not being actively developed right now, because we don't at the moment have a roadmap to get content into Markdown in GitHub.

This is distinct from the linter being discussed in this thread, which lints MDN Wiki content, and which *is* under active development.

Issue #392, opened 2020-04-09
U - Test ingredient and recipe linting


ddbeck:
As an editor of wiki content, I want to be able to verify and demonstrate linting behavior independently of the wiki itself (e.g., for debugging and CI). We should have a test suite for `scraper-ng` that doesn't rely on the wiki being available or live wiki pages being in a specific state.

## Acceptance criteria 

- [x] Implement one or more tests for `scraper-ng`
- [x] Add tests to CI

wbamberg:
@ddbeck , I think we can close this now, can't we?

ddbeck:
Yes, agreed!

PR #395, opened 2020-04-16
Demonstrating tests for wiki linting

ddbeck:
This PR shows off two potential ways of testing the linter for #392. Both use [Jest](https://jestjs.io/) to run the tests and provide assertions, but the tests themselves show two different ways of structuring our tests. There are benefits and disadvantages to each. Let's take a look!

## General notes

To run the tests, check out the branch, run `npm install`, and, finally, `npx jest`.

I've set up testing with Jest here because it's very popular and it's used elsewhere in the MDN code base (in KumaScript, in particular). Apart from my reservations about using a project from Facebook, it's very pleasant to use. Running tests gives pretty looking results, writing assertions isn't too opaque, and there's a nice VS Code plugin that automatically reruns tests as type.

In all, I went from no tests to a small suite in about 2 hours—I was sure I was going to spend all day on this. I imagine I could write tests for all our implemented ingredients in less than a day's work.

## Unit-style tests

`data-browser-compatability.test.js` shows how we could test ingredient handlers directly. A nice thing about this style of testing is that it's very self-contained. It bypasses all the file handling stuff and goes directly to the question that the code actually answers: does some tree produce errors and, if so, which ones? The downside is that you have to do a bit more work to set up a test (composing a tree with [hastscript](https://github.com/syntax-tree/hastscript)). These tests also won't validate anything about recipes. That'll require something else.

## Integration-style tests

`test/test.js` goes after everything at once: you provide some source HTML and a recipe path. The nice thing about this is that it runs a chunk of sample HTML and a recipe through the complete wringer. You don't have to think too hard about how to write a test case (you can copy source directly from the wiki) or how it relates to the linter output we'll actual use.

The not so nice thing is that it runs it through the complete wringer. If, for example, you copy some source from the wiki to test an ingredient, you'll also get warnings about disallowed macro calls and such. This means that changes to different parts of the linter may lead to fiddly changes in unrelated test cases, or you'll have to do some deep inspection on file messages to write tests narrowly.

This implementation isn't very ergonomic just yet. It'd be nice to get test HTML source from files instead of strings or recipes from objects (instead of YAML files) and so on. That'd require a bit of helper code (and some changes to the way recipes are processed normally), which I decided not to write unless I had some validation first. Not hard to do by any means, but it's time that I didn't want to spend ahead of need.

## Feedback wanted

The main things I'm looking for feedback on here are:

* Can you follow the tests? If not, which ones?
* Can you imagine writing tests like this? Are there things which are appealing or unappealing about writing such tests?
* Does this increase you confidence in our ability to lint wiki pages?

That said, I'm happy to take comments on anything.


wbamberg:
This is really nice @ddbeck !

I like the integration-style tests better, for a few reasons:

* I don't have to learn a new syntax for describing markup
* I like the tests living in a separate place from the linter code (I guess you could do this with the unit test style, but it might be awkward)
* I like the fact that integration-style tests don't need to know the internal details of the linter code: they can just call a top-level function that's the same for all tests. Like this you need different framework code for calling ingredient handlers, or calling the test for unknown H2, or... and when we change the internal implementation, we have to update the tests.

But also I guess the question is: what are our testing requirements? We could potentially write a lot of tests for the linter, but is it a good use of our time? What tests give us the most value?

Perhaps we could have a few tests for each recipe?

> Can you follow the tests?

Yes.

> Can you imagine writing tests like this? Are there things which are appealing or unappealing about writing such tests?

Yes. As I said I think the integration-style will be easier to write, and I like the API you have there where we pass a blob of HTML and a recipe in, and look at the messages coming out.

I think it would definitely be good to be able to supply the HTMLs as files, and also to be able to provide "expected messages" as data rather than code. So if a test author could provide:
* an HTML blob
* a recipe identifier
* some JSON describing expected messages

...and have the test framework run that, then this seems like it might be quite nice.

> Does this increase you confidence in our ability to lint wiki pages?

Yeah, I think so. So for instance when I was writing the linter for class members I'd go and edit https://wiki.developer.mozilla.org/en-US/docs/User:wbamberg/test-linter2 to have certain content, then check the messages. It would have been better to have done this by writing a few tests (and it means they continue to work as regression tests if I have to refactor the linter).

ddbeck:
Thank you, Will. This is very helpful feedback. I've got some specific responses below. But overall, I'm going to go in the direction you suggest, with the integration style tests. With some changes to make recipe handling easier, I think it would be relatively easy to write ingredient specific tests as well.

> I like the API you have there where we pass a blob of HTML and a recipe in, and look at the messages coming out

Yeah, I preferred this API as well. After opening the PR, I had a closer look at the Jest docs and I found it was rather pleasant to extend the Jest API too, to make looking at the messages a bit easier to write.

> some JSON describing expected messages

I like the idea of wholly data-driven tests too, though the messages part is a bit complicated. It'd be fairly easy to verify snapshots of messages from JSON, but these might be brittle tests. I'll give this a try, but if it gets complicated, I may defer this to a follow up PR.

ddbeck:
OK, I've pushed a bunch of changes to make the integration tests work a bit better. I may have gotten over enthusiastic. Some highlights:

* df762a2f443886aed76ef385994ffe78143e61e2 Removed the unit tests and `hastscript`.
* 0da8ffae8b4c9ef013f40e04a4fe92831c7434ad I used Jest's `expect.extend` machinery to add a custom assertion (in this case, that at least one message on a file has a `ruleId` that matches some string or regex). Now you can require a message with `expect(file).hasMessageWithId(someStringOrRegex)`.
* e26fb51c64af617eacc0b35c88d1229ca6566516 A hopefully modest refactor to loading recipes. This allows tests to insert recipes directly, a handy convenience for writing ingredient-specific tests (as in 6b0548edbc8f144fac41ca99d1494c5b10a638bd). I think there are some more improvements to be made here, but I'm going to save it for another time.
* f098f036726175d39e4a946053ad7eb67f8ec2da Tests that every required recipe ingredient triggers a message against an empty page. This test verifies that _something_ happens when you require an ingredient (it'll be more meaningful once we've implemented handlers for every ingredient).
* 10607875b5fc366efd44af6c1ef45ac68de7ca81 Now that there are multiple test files, I've set up Jest's config. Our monopackage approach might be getting a bit strained. Oh well.
* 3f14efee800b60eef6dc2e507aff6c0595b1db81 And the payoff for all that setup: a rather tidy test against a whole page from MDN, read from a file. 🎉 

Taking this back out of draft mode, so I will quit doing things. 😄

scripts/scraper-ng/test/javascript-constructor.test.js
```diff
@@ -0,0 +1,19 @@
+const fs = require("fs");
+const path = require("path");
+
+const { processFromSource, recipePath } = require("./utils");
+
+test.only("javascript-constructor", () => {
+  const file = processFromSource(
```
wbamberg, line 7: If we think it is a common pattern to supply:
* HTML in a file
* a built-in recipe

...then we can make this a bit nicer for authors, by supporting `processFromFile` than takes a path, and by letting the recipe argument be any one of:
* recipe YAML
* a path to a YAML file (so we can supply custom recipes, but from a file)
* the name of a built-in recipe

Then this could be:

```js
const file = processFromFile(
  path.resolve(__dirname, "./javascript-constructor.html"),
  "javascript-constructor"
);
```

...which is easier to understand.

ddbeck: Quite right. See 54c6d94f1536b9832f407843d3dc80b521f9472c for the refactor.

ddbeck:
Thanks, Will. Some good suggestions there. I've made the API for processing a page a bit nicer and reorganized the test folder.

One more comment:

> I'm still not sure which tests exactly we'll want to have, and looking at this now, the "ingredient-data.browser_compatibility.test.js" looks more useful than the "javascript-constructor.test.js" one. If I were writing tests for the data.examples linter, for instance, it obviously seems more useful to clone "ingredient-data.browser_compatibility.test.js". But I think we can totally merge this and then see how we want to use it.

Yeah, I see this too. That said, `javascript-constructor.test.js` is basically the start of a couple variations of tests that I want to have to lint ingredient order. I imagine we'll have different test styles depending on the scope of the changes we need to verify (whether we're linting whole pages, specific ingredients, or arbitrary fragments of HTML).

ddbeck:
> I think at some point it would be great to document how to write a test, but I think before that we might want to organize the scraper-ng into a proper structure maybe? (I still sometimes `cd` into the wrong scraper and it might be hard to tell from the top-level repo that scraper-ng is actually nowadays the heart of this whole project?). Not a requirement for this PR of course, but just a thought while reviewing this from a birds eye perspective.

Yeah, our repo layout has become a little bit awkward. I had to do some slightly odd things with jest's config, for example, because everything's in one big package. It might be good (and it might make Peter B happy) if we set this up as a yarn workspace, now that I think about it.

I just pushed a file with some initial documentation on testing. Hopefully it's useful!

PR #399, opened 2020-04-20
Rename 'data.class_constructor' to 'data.constructor'

wbamberg:
As discussed in https://github.com/mdn/stumptown-content/pull/397, I think we ought to use `data.constructor` to name the "Constructor" ingredient in JavaScript class pages (i.e. to name the page element at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Constructor).


Elchi3 (approved):
Makes sense to me, thanks Will!

Issue #350, opened 2020-03-09
U - Write a linter ingredient handler for data.examples


wbamberg:
The linter currently does check for the existence of an "Examples" section, but stumptown expects examples to have a particular structure, and the linter does not check that the content of the "Examples" section is in the right structure.

## Acceptance criteria

The linter checks the correctness of the `data.examples` ingredient in recipes that include it, according to the specification agreed in https://github.com/mdn/sprints/issues/3075.

wbamberg:
I'd like us to consider whether the [structure for examples described in the specification doc](https://github.com/mdn/stumptown-content/blob/master/project-docs/javascript-linter-spec.md#dataexamples) is unecessarily strict.

In the specification, everything that isn't actual example code (including explanatory prose) has to be corralled into the start of the example, and there may be only one code block (for each language) containing example code.

Looking through some JS pages, I think many examples will fail this test. For example, it means you can't intersperse prose and code, as in this example for [`String.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice#Using_slice_with_negative_indexes).

**For examples that aren't live samples**, there's no functional reason to impose these kinds of limitations. Meaning, there's nothing about them that is needed by structured Markdown as we have envisaged it. We should still I think mandate that each example lives in H3 sections with titles, but apart from that we could just treat each example as unstructured prose. So here the key question seems to be: **is there any non-functional reason to impose these restrictions?** Meaning, would the added consistency help users, or would these restrictions make the examples worse, by restricting writers too much?

My inclination at this point is that we should relax these restrictions, and say:

* for all examples, each example must live in its own H3 section
* for non-live samples, there are no particular restrictions beyond that

**For examples that are live samples**, I think the situation is a bit different. I would like to avoid the KS-style approach where we have to embed directives in the example to say where to put the output iframe, and I would like to avoid having lots of extra markup in the example to say which bits of code are included in which live example. So I would suggest we could say:
* a single example may only contain one live sample
* the output of the live sample must always appear at the end of the example, after all the prose and sample code
* all code samples in an example section must be part of the live sample (we could fairly easily drop this restriction, and have writers add a class to the "code fence" for code samples to indicate that a particular code block is part of the live sample. But I think it would make things more complicated to maintain). 

But in this model it would still be possible to intersperse code and prose, if people want to do that.

[...]

PR #401, opened 2020-04-22
Add ingredient handler for 'data.examples'

wbamberg:
This extends the linter to lint examples. It uses the spec described in https://github.com/mdn/stumptown-content/issues/350#issuecomment-617455037, not the one currently in the [doc](https://github.com/mdn/stumptown-content/blob/master/project-docs/javascript-linter-spec.md). So merging it is contingent on agreeing that change to the spec, and updating the doc.

If the testing PR lands before this, I will add some tests as well :).

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-examples.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-examples.js
@@ -83,7 +83,7 @@ function handleDataExamples(tree, logger) {
       (node) => node.tagName === "h3",
       examplesSection
     );
-    checkExample(body, exampleSection, logger);
+    checkExample(exampleSection, logger);
   }
 }
 
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-examples.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-examples.js
@@ -1,0 +1,90 @@
+const select = require("hast-util-select");
+const visit = require("unist-util-visit");
+
+const utils = require("./utils.js");
+
+/**
+ * Function returning true only if the given node is a text node
+ * that contains only newlines.
+ */
+function isNewlineOnlyTextNode(node) {
+  return node.type === "text" && node.value.match(/^\n*$/);
+}
+
+/**
+ * Checks a single example. This only does any checking for live samples:
+ * that is, examples that contain a call to the EmbedLiveSample macro.
+ */
+function checkExample(body, exampleNodes, logger) {
+  let liveSampleNode = null;
+  visit(exampleNodes, (node) => {
+    // If we have already seen an EmbedLiveSample call...
+    if (liveSampleNode !== null) {
+      // ...then the only nodes allowed after that point are text nodes containing only newlines
+      if (!isNewlineOnlyTextNode(node)) {
+        logger.fail(
+          node,
+          "EmbedLiveSample must be final node",
+          "nodes-after-live-sample"
+        );
+      }
+    }
+    if (utils.isMacro(node, "EmbedLiveSample")) {
+      liveSampleNode = node;
+    }
+  });
+  // If this example didn't contain a call to EmbedLiveSample, there are no further checks
+  if (!liveSampleNode) {
+    return;
+  }
+  // The first child of `exampleNodes` is the title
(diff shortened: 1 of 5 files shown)


wbamberg:
I've added some commits:
* https://github.com/mdn/stumptown-content/pull/401/commits/0a241b34178ed3f8aae4394a2e1eba27d8209a23 : added some tests
* https://github.com/mdn/stumptown-content/pull/401/commits/6a078aeeaf120095f5a4cac7a1f86b47af8b59fd: updated the specification according to the discussion here: https://github.com/mdn/stumptown-content/issues/350#issuecomment-617455037

scripts/scraper-ng/test/ingredient-data.examples.test.js
```diff
@@ -0,0 +1,107 @@
...
+  <pre class="brush: html">&lt;h1&gt;A heading&lt;/h1&gt;</pre>
+  <pre class="brush: css">h1: { color: red; }</pre>
+  {{EmbedLiveSample("Another_example")}}`,
+
+  valid_simple_live_mixture: `<h2 id="Examples">Examples</h2>
+<h3 id="An_example">An example</h3>
+  <p>Some prose description</p>
+  <pre class="brush: html">&lt;h1&gt;A heading&lt;/h1&gt;</pre>
+  <pre class="brush: css">h1: { color: blue; }</pre>
+  {{EmbedLiveSample("An_example")}}
```
ddbeck, line 28: Maybe we could have some newlines after this macro call, since the spec mentions this as a possibility

```suggestion
  {{EmbedLiveSample("An_example")}}
  
  
```

scripts/scraper-ng/test/ingredient-data.examples.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+    expect(file.messages.length).toBe(1);
+    expect(file).hasMessageWithId(/embedlivesample/);
+  });
+
+  test("missing Examples H3", () => {
+    const file = process(sources.missing_example_h3, testRecipe);
+
+    expect(file.messages.length).toBe(1);
+    expect(file).hasMessageWithId(/data.examples\/missing-example-h3/);
```
ddbeck, line 82: For future reference, `hasMessageWithId` also matches partially with a string instead of a regex, if you don't want to fuss with escaping slashes. Not necessary to change here though, since it works and it's still pretty readable.

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-examples.js
```diff
@@ -0,0 +1,90 @@
...
+ */
+function isNewlineOnlyTextNode(node) {
+  return node.type === "text" && node.value.match(/^\n*$/);
+}
+
+/**
+ * Checks a single example. This only does any checking for live samples:
+ * that is, examples that contain a call to the EmbedLiveSample macro.
+ */
+function checkExample(body, exampleNodes, logger) {
```
ddbeck, line 18: I think `body` is unused here

```suggestion
function checkExample(exampleNodes, logger) {
```

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-examples.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  for (const exampleTitle of exampleTitles) {
+    // Extract and check each example
+    const exampleSection = utils.sliceBetween(
+      exampleTitle,
+      (node) => node.tagName === "h3",
+      examplesSection
+    );
+    checkExample(body, exampleSection, logger);
```
ddbeck, line 86:
```suggestion
    checkExample(exampleSection, logger);
```

ddbeck (changes requested):
Hey Will, this looks great. I had just a few basically nit picky suggestions.

Also, I'm really happy with how much code gets reused here and how the tests help illustrate what the code does—it's almost as if we're making useful building blocks for making sense of our content. 😂

scripts/scraper-ng/test/ingredient-data.examples.test.js
wbamberg: Oh, I somehow didn't see the suggested change, and just added this independently.

wbamberg:
Thanks Daniel! It was nice writing this linting check and even more so I think to write tests for the linter. Great work getting the testing framework done.

ddbeck (approved):
🎉

Issue #332, opened 2020-03-03
Write linter ingredient handler for data.info_box


ddbeck:
A follow up to #320

We need an ingredient handler for `data.info_box`. This seems sort of straight forward—look for a table in some particular location—but info boxes seem to vary significantly across page types. For example, CSS pages use `cssinfo` macros in a different place than JS pages use `js_property_attributes`.

I would expect whoever takes this on to do one of:

* make a plan for distinguishing different info boxes (if not actually implementing all of them)
* break info boxes into different domain-specific types (e.g., `data.js_info_box`)
* get it working for JS and kick the can down the road for other content areas

Elchi3:
This is one of the top warnings right now:

---
99 warnings    javascript-property/data.info_box/handler-not-implemented

wbamberg:
I wrote this in an email (April 9th, when we were sprint planning) but unfortunately forgot to update the other places:

> I've not included this in the spec and suggest we don't do data.info_box for now. It's a rather controversial topic that would take a lot of modelling work, and isn't needed for the JS docs to be "stumptown compatible" (scrapable) unless we want to structure that content.

Unless anyone objects I suggest we resolve this by removing `data.info_box` from the recipe.

Elchi3:
Closing this then.

PR #415, opened 2020-04-30
Remove data.info_box

wbamberg:
See https://github.com/mdn/stumptown-content/issues/332#issuecomment-622056087.

We haven't specced linter behavior for the `data.info_box` ingredient of JS properties, because I was thinking we should just treat it as prose for now.

We don't at the moment understand quite how we should [represent "info boxes" in MDN structured content](https://docs.google.com/document/d/1fwbTs6uoeAfGjjZef3cOOzG1Bx-S8sbLbCBjzx9gBPY/edit#). So representing them would involve us in a substantial amount of modelling work. Unlike the CSS info boxes, the boxes in JS property pages are simple tables that don't use any external data source, so we don't _have_ to do anything special (i.e. non-prose-y) with them even when scraping them into Markdown. We only have to do anything special with them if we want to structure the data they contain, and it's not obvious to me that doing that adds a lot of value. Even if it is worth it, we can defer that extra "deep structure" until we want it.

But we (I) forgot to remove the ingredient from the recipe, so that's what this PR does.

### DIFF
--- a/recipes/javascript-property.yaml
+++ b/recipes/javascript-property.yaml
@@ -7,7 +7,6 @@ body:
   - prose.short_description
   - data.interactive_example?
   - prose.description?
-  - data.info_box
   - prose.*
   - data.examples
   - data.specifications


Elchi3 (approved):
Thanks Will, I agree to remove it at this stage.

Issue #407, opened 2020-04-30
Fix linting errors "dt elements in link lists must contain a single anchor element"


Elchi3:
432 errors javascript-class/data.instance_methods
129 errors javascript-class/data.instance_properties
69 errors   javascript-class/data.static_methods
49 errors   javascript-class/data.static_properties
38 errors   javascript-namespace/data.static_methods
9 errors     javascript-namespace/data.static_properties
8 errors     javascript-namespace/data.constructor_properties

= 734 errors and thus the majority of all errors we currently have when linting all of the JS MDN docs.

I believe we can fix this by either allowing or by removing `jsxref` macros in `<dt>`

wbamberg:
-> https://github.com/mdn/stumptown-content/pull/419: we should now tolerate `jsxref` in lnk lists.

Elchi3:
Thanks Will! This reduced the error amount drastically. Remaining:


---
52 errors      javascript-class/data.instance_properties?/only-single-anchor-element-or-xref-in-link-list-dt
3 errors       javascript-class/data.instance_methods?/only-single-anchor-element-or-xref-in-link-list-dt
1 error        javascript-class/data.static_methods?/only-single-anchor-element-or-xref-in-link-list-dt

Elchi3:
@wbamberg what did we decide about readonly inline banners? The remaining failures seem to be about that. See e.g. https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float32Array#Instance_properties

wbamberg:
We talked about this here: https://github.com/mdn/sprints/issues/2520.

`readonlyinline` is mostly problematic to remove when it's not attached to a link. When a badge is attached to a link, we think it is OK to remove, because we think the linked page will tell the reader that the thing is read only (or experimental, or whatever), via a banner in the target page.

But by definition, in "link lists" the badge will always be attached to a link. And in the example you gave of https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float32Array#Instance_properties, the linked pages do indicate that the property is read only (although not using a banner).

In the page you linked to, the `dd` content also says that the property is read only, so I think it would be fine to remove the macro calls (in fact it looks weird having both).

Elchi3:
I've removed the macro calls (also given its rendering is broken) and added "ready only" in the dd text if it wasn't already there.

Some other fixes were needed but I have now cleared all errors of this kind.

PR #419, opened 2020-05-01
Accept jsxref as dt element contents inside link lists

wbamberg:
See https://github.com/mdn/stumptown-content/issues/407.

This fixes the linter errors by tolerating `{{jsxref}}` in `dt` elements in link lists.

I'm still not certain that this is preferable to rendering them out, but what tipped me into doing this is that our spec says we should accept `{{jsxref}}`: https://github.com/mdn/stumptown-content/blob/master/project-docs/javascript-linter-spec.md#dataconstructor_properties.

Since this didn't have any tests I also added some.

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/link-list-checker.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/link-list-checker.js
@@ -2,6 +2,21 @@ const select = require("hast-util-select");
 
 const utils = require("./utils.js");
 
+/**
+ * Returns true only if the given node contains a single child,
+ * andthe child is either an A element or a call to the jsxref macro.
+ */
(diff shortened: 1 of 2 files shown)


scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/link-list-checker.js
```diff
@@ -2,6 +2,21 @@ const select = require("hast-util-select");
 
 const utils = require("./utils.js");
 
+/**
+ * Returns true only if the given node contains a single child,
+ * andthe child is either an A element or a call to the jsxref macro.
```
ddbeck, line 7:
```suggestion
 * and the child is either an A element or a call to the jsxref macro.
```

ddbeck (approved):
This looks good. Suggested a change to a tiny typo, but it can merge with or without that change at your convenience.

PR #421, opened 2020-05-01
Add handler for the data.constructor ingredient

wbamberg:
This add a handler for the `data.constructor` ingredient, according to the spec: https://github.com/mdn/stumptown-content/blob/master/project-docs/javascript-linter-spec.md#dataconstructor - well, mostly. I didn't check the value of `NameOfTheObject`, since that looks hard to get, and I thought this would be good enough.

I've deferred some of the checking to the generic link list checker, and added the extra custom checks after that.

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
@@ -69,7 +69,7 @@ function handleDataConstructor(tree, logger) {
   if (!checkTag(dd.children[1], "a")) {
     logger.fail(
       section,
-      "Constructor description second node is incorrect",
+      "Constructor description must contain a link after `Creates a new`",
       "constructor-description-second-node"
     );
     ok = false;
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
@@ -60,7 +60,7 @@ function handleDataConstructor(tree, logger) {
   if (!checkText(dd.children[0], "Creates a new ")) {
     logger.fail(
       section,
-      "Constructor description first node is incorrect",
+      "Constructor description must be in the form `Creates a new <a>...</a> object.`",
       "constructor-description-first-node"
     );
     ok = false;
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
@@ -51,7 +51,7 @@ function handleDataConstructor(tree, logger) {
   if (dd.children.length !== 3) {
     logger.fail(
       section,
-      "Constructor description must contain three nodes",
+      "Constructor description must be in the form `Creates a new <a>...</a> object.`",
       "constructor-description-three-nodes"
     );
     return null;
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
@@ -1,0 +1,84 @@
+const select = require("hast-util-select");
+
+const utils = require("./utils.js");
+const checkLinkList = require("./link-list-checker.js");
+
+function checkText(node, text) {
+  return node.type === "text" && node.value === text;
+}
+
+function checkTag(node, tag) {
+  return node.type === "element" && node.tagName === tag;
+}
+
+/**
+ * Handler for the `data.constructor` ingredient.
+ */
+function handleDataConstructor(tree, logger) {
+  // Check common link list structure
+  checkLinkList("Constructor", tree, logger);
+
+  // Constructor is a mandatory property
+  const heading = select.select(`h2#Constructor`, tree);
+  if (heading === null) {
+    logger.expected(tree, `h2#Constructor`, "expected-heading");
+    return;
+  }
+
+  // This link list is only allowed one entry
+  const section = utils.sliceSection(heading, tree);
+  const dts = select.selectAll("dt", section);
+  if (dts.length !== 1) {
+    logger.fail(
+      section,
+      "Constructor section may only contain one DT item",
+      "only-single-constructor-dt"
+    );
+  }
+  const dds = select.selectAll("dd", section);
+  if (dds.length !== 1) {
+    logger.fail(
+      section,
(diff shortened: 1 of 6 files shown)


wbamberg:
I just added a couple of commits:

* https://github.com/mdn/stumptown-content/pull/421/commits/84ca61f6ee5ec326e8464f7a829f1fc078a928df : makes the handler return a the node it found
* https://github.com/mdn/stumptown-content/pull/421/commits/a80d8ea18bb62c1de6969c7c93a061905408d6da : updated tests to test the returned position, along the lines already added in the lint-order PR. I ended up factoring out a few test utility functions there too.

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
```diff
@@ -0,0 +1,94 @@
...
+    );
+    return null;
+  }
+
+  // The dd must be of the form: "Creates a new <a>...</a> object."
+  const dd = dds[0];
+  if (dd.children.length !== 3) {
+    logger.fail(
+      section,
+      "Constructor description must contain three nodes",
```
ddbeck, line 54: The messages here are kinda particular to the way hast represents the HTML. I'm going to suggest some alternatives

```suggestion
      "Constructor description must be in the form `Creates a new <a>...</a> object.`",
```

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
```diff
@@ -0,0 +1,94 @@
...
+      "only-single-constructor-dd"
+    );
+    return null;
+  }
+
+  // The dd must be of the form: "Creates a new <a>...</a> object."
+  const dd = dds[0];
+  if (dd.children.length !== 3) {
+    logger.fail(
+      section,
```
ddbeck, line 53: Target the offending node more closely:

```suggestion
      dd,
```

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
```diff
@@ -0,0 +1,94 @@
...
+      "Constructor description must contain three nodes",
+      "constructor-description-three-nodes"
+    );
+    return null;
+  }
+
+  if (!checkText(dd.children[0], "Creates a new ")) {
+    logger.fail(
+      section,
+      "Constructor description first node is incorrect",
```
ddbeck, line 63:
```suggestion
      "Constructor description must be in the form `Creates a new <a>...</a> object.`",
```

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
```diff
@@ -0,0 +1,94 @@
...
+      "constructor-description-first-node"
+    );
+    ok = false;
+  }
+
+  if (!checkTag(dd.children[1], "a")) {
+    logger.fail(
+      section,
+      "Constructor description second node is incorrect",
```
ddbeck, line 72:
```suggestion
      "Constructor description must contain a link after `Creates a new`",
```

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
```diff
@@ -0,0 +1,94 @@
...
+      "constructor-description-second-node"
+    );
+    ok = false;
+  }
+
+  if (!checkText(dd.children[2], " object.")) {
+    logger.fail(
+      section,
+      "Constructor description third node is incorrect",
```
ddbeck, line 81:
```suggestion
      "Constructor description must end with " object.",
```

scripts/scraper-ng/test/framework/utils.js
```diff
@@ -66,7 +66,57 @@ function recipePath(recipeName) {
   return path.join(recipesDir, recipeName + ".yaml");
 }
 
+/**
```
ddbeck, line 69: If you're feeling ambitious, this could be done as a custom matcher for Jest in [jest.setup.js](https://github.com/mdn/stumptown-content/blob/master/scripts/scraper-ng/test/framework/jest.setup.js). The Jest matcher API is [a little bit complex](https://jestjs.io/docs/en/expect#custom-matchers-api) though, so it could happen in a separate PR

scripts/scraper-ng/test/framework/utils.js
```diff
@@ -66,7 +66,57 @@ function recipePath(recipeName) {
...
+ * Given an ingredient, an ingredient name and a tag name, check that:
+ * * the ingredient exists
+ * * the ingredient name matches the name given
+ * * the ingredient's position node is an element and its tag matches the given tag.
+ *
+ * @param {String} ingredient - an ingredient including a name and a position node
```
ddbeck, line 75: Isn't `ingredient` supposed to be an object? This comes up throughout, so I wasn't completely sure

scripts/scraper-ng/test/framework/utils.js
```diff
@@ -66,7 +66,57 @@ function recipePath(recipeName) {
   return path.join(recipesDir, recipeName + ".yaml");
 }
 
+/**
+ * Given an ingredient, an ingredient name and a tag name, check that:
```
ddbeck, line 70: Perhaps this could be a little more clear about what's meant as "an ingredient" here (e.g., a `file.data.ingredient` object)?

ddbeck (changes requested):
Thanks for this, Will! It looks really good. I've made some line comments to improve documentation and descriptiveness, but I don't think there are any functional issues here.

scripts/scraper-ng/test/framework/utils.js
wbamberg: I did think about this, but... the code to do this is quite a lot more complicated and it wasn't at all obvious to me that it was worth it.

scripts/scraper-ng/test/framework/utils.js
wbamberg: This is a bit tricky because `file.data.ingredient` doesn't actually exist (it's an array of `file.data.ingredients` objects). It seemed more helpful to describe what they are, but extending the description in this bit adds a lot of repetition with the `@param` line, which just looks silly. So I opted to keep this short, and extended the `@param` line a bit.

ddbeck: Ah, got it. This makes sense.

scripts/scraper-ng/test/framework/utils.js
ddbeck: Fair enough 😃

ddbeck (approved):
Looks good. Thank you, Will! 🎉

PR #424, opened 2020-05-04
Make markdown linter match files better.

jpmedley:
Addresses #375


wbamberg:
@jpmedley , it looks like Prettier is failing the build. You can reformat by running:

> npm run prettier-format

PR #431, opened 2020-05-08
Update ctor handler to look for <code> instead of <a>

Elchi3:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Constructor

The `<dd>` should be "Creates a new `BigInt` object." but `BigInt` shouldn't link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt because that's the same page.

Another thing I wanted to fix, but I don't know how is that the end of the `<dd>` shouldn't be required to be "object.". Imo, we need more freedom there, because sometimes a second sentence follows, see e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Constructor (and a handful more, these are the last remaining linter errors of this category).

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
@@ -46,12 +46,12 @@ function handleDataConstructor(tree, logger) {
     return null;
   }
 
-  // The dd must be of the form: "Creates a new <a>...</a> object."
+  // The dd must be of the form: "Creates a new <code>...</code> object."
   const dd = dds[0];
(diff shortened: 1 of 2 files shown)


wbamberg (approved):
Thanks @Elchi3 !

wbamberg:
> Another thing I wanted to fix, but I don't know how is that the end of the <dd> shouldn't be required to be "object.". Imo, we need more freedom there, because sometimes a second sentence follows, see e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Constructor (and a handful more, these are the last remaining linter errors of this category).

We could just say it has to start with " object."? It does seem reasonable to relax this.

PR #433, opened 2020-05-09
Relax ctor <dd> linting

Elchi3:
See https://github.com/mdn/stumptown-content/pull/431#issuecomment-626029698


wbamberg:
Thanks @Elchi3 ! Is there a reason to use `includes()` rather than `startsWith()`? It seems the latter is what we actually want to enforce, and what the error message says:

> Constructor description must end first sentence with ' object."

With this PR a description like:

    Creates a new <code>Thing</code>, which is a banana. Not an object.

would pass.

Elchi3:
Thanks for catching that, Will! Updated to use `startsWith`.

PR #438, opened 2020-05-13
Lint prose.* order

### LINKED ISSUE #338: U - Lint ingredient order
A follow up to #320 .

We should lint for the order of ingredients in pages.

## Implementation notes

This is a little bit tricky. Some ingredients are only meaningful in order (e.g., `prose.short_description`) and linting the ingredient in-depth is the only meaningful check on it. But others can exist anywhere in a page (most prose sections).

A possible route to implementation might be to use `html-require-recipe-ingredients`. After linting ingredients, it could check to see if any new messages were logged to a file. If not (i.e., if all ingredients matched), then it could check the order of section headings. This might be good enough for many, maybe even most, pages.

## Acceptance criteria

The linter is able to check that all ingredients appear in the correct order, according to the specification agreed in https://github.com/mdn/sprints/issues/3075.

ddbeck:
This PR adds a handler (of sorts) for `prose.*` ingredients, which allows for the linting of `prose.*` headings. With this PR, all ingredients can be linted for order. This fixes #338.

## Notes

* I was really glad we had the stumptown call this week. It gave me an idea of how to structure this such that none of the existing tests required changes. 🎉
* The diff is not quite as clean I would've liked. I got to my solution a bit circuitously, so there are a few lines changed that aren't really germane. I can revert them, if they're distracting.
* Unlike the previous PR for linting order, this one actually helped me discover some pages with problems, such as [parseInt](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) (where the `prose.*` sections appear after examples, instead of before) or [AggregateError](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) (which can't decide if it's class page or a constructor page).
* Possible follow up: For all the specified ingredients, we now have handlers. If we want to, we can throw proper exceptions for undhandled ingredients. This is probably a good idea, but it requires reworking a few tests that didn't seem relevant here, or dumping our old HTML or CSS recipes (which would impact other things and be controversial). In any case, I left a TODO comment about it.

### DIFF
--- a/scripts/scraper-ng/rules/html-require-ingredient-order.js
+++ b/scripts/scraper-ng/rules/html-require-ingredient-order.js
@@ -83,42 +83,48 @@ function attacher() {
   };
 }
 
-const ingredientsWithoutDefinedH2s = [
+// Ingredients that do not have a corresponding H2 element
+const ingredientsWithoutH2s = [
   "prose.*",
   "prose.short_descriptions",
   "data.interactive_example",
 ];
 
+/**
+ * Collect H2s not otherwise expected in an ingredient.
+ *
+ * @param {Object} tree - A hast tree
+ * @param {vfile} file - A vfile with a `file.data.ingredients` array and a `file.data.recipe` object
+ * @returns {Array} an array of `file.data.ingredients` member-like objects with `name` and `position` properties
+ */
 function collectProseStar(tree, file) {
-  const reservedIds = [];
-  for (const ingredient of file.data.recipe.body) {
-    if (!ingredientsWithoutDefinedH2s.includes(ingredient)) {
-      const { name } = ingredient.match(/[.](?<name>\w+)/).groups;
-      const id = `#${name.charAt(0).toUpperCase() + name.slice(1)}`;
-      reservedIds.push(id);
-    }
-  }
+  const reservedIds = file.data.recipe.body
+    .filter((i) => !ingredientsWithoutH2s.includes(i))
+    .map(ingredientToSelector);
 
-  const h2s = selectAll(`h2:not(${reservedIds.join()})`, tree);
-  const ingredientNodes = file.data.ingredients.reduce((arr, curr) => {
-    if (curr.position !== null) {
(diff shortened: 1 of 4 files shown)


ddbeck:
👋 Hey folks, just a reminder that this is still awaiting review

scripts/scraper-ng/rules/html-require-ingredient-order.js
```diff
@@ -4,29 +4,49 @@ const source = "html-require-ingredient-order";
...
     );
 
-    for (let i = 0; i < pageIngredients.length; i++) {
-      const currentIngredient = pageIngredients[i];
-      const expectedIngredient = recipeIngredientNames[i];
+    // Pad the sequence of expected ingredients with additional `prose.*`
+    // entries to cover all such ingredients
+    if (recipeIngredientNames.includes("prose.*")) {
+      const padCount = pageIngredients.reduce(
+        (sum, ingredient) => sum + (ingredient.name == "prose.*" ? 1 : 0),
```
wbamberg, line 29:
```suggestion
        (sum, ingredient) => sum + (ingredient.name === "prose.*" ? 1 : 0),
```

ddbeck:
> So we could change our logging policy, and just log one error as soon as an out-of-order ingredient is found, then stop.

This is a pretty good idea and one we hadn't really talked about over email yet. There's something to the idea that a _page_ is disordered when any one ingredient is out of order and that it's a failure of the page, rather than of any specific ingredient.

On this point specifically, I'd like to see how this sorting check performs when actually cleaning up pages. If the messages we've got now are too complex or has bugs, then I'd like to try the one-failure-and-stop approach.

ddbeck:
@Elchi3 if you plan to review this, let me know. Otherwise, I'll merge this tomorrow if there's no further commentary. 

(Sorry for the notification noise. Forgot about this.)

ddbeck:
Excellent. Thank you both for the feedback.

Issue #440, opened 2020-05-14
U- Write recipe for CSS selectors


wbamberg:
This issue is split out of https://github.com/mdn/stumptown-content/issues/396, which wants us to write recipes for all the CSS reference documentation page types.

I thought to keep things a bit easier to follow it might make sense to have issues to track different CSS recipes. This one is for **CSS selectors**. 

# Acceptance Criteria 
- [ ] recipe for CSS selector pages is written and checked into stumptown-content
- [ ] spec for how to lint with this recipe is written and checked into stumptown-content

wbamberg:
# CSS Selectors (pseudoclasses and pseudoelements)

I think there are about 137 pages here. 71 are prefixed (40 moz, 14 ms, 17 webkit).

The unprefixed ones have a fairly consistent structure:

* short description
* static code example (see for example  [`::grammar-error`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::grammar-error))
* optional more prose
* H2#Syntax
* H2#Allowable/permitted properties (optional)
* H2#Accessibility concerns (optional)
* H2#Examples
* H2#Specifications
* H2#Browser compatibility
* H2#See also

A few things worth considering:

## The static code example at the start

These pages don't have interactive examples, but generally have a static code block instead, in the same place. So options here are:

1. Don't structure/enforce this, but just treat the static code example as part of a prose H2#Description section.
2. Don't structure/enforce this, but just treat the static code example as part of a prose H2#Syntax section (if we treat that as prose rather than formal syntax, see below).
3. Invent a new ingredient that's a static code sample.

I think I'd favour (2).

## H2#Syntax

* This usually contains the formal syntax via a call to `{{csssyntax}}`, and only the formal syntax (i.e. no prose description of the syntax, no `{{cssinfo}}`. 

* Occasionally it does not use the formal syntax (e.g. [`::grammar-error`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::grammar-error)).

* Occasionally it uses the formal syntax and a prose description (e.g. [`:dir`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/:dir#Syntax) ).

* The formal syntax is almost always extremely simple (e.g. [`:active`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/:active#Syntax)).

[...]

Issue #396, opened 2020-04-17
U - Write recipes for CSS properties


wbamberg:
**Note**: this story was originally to:

(1) identify which distinct recipes we'd need to write for CSS
(2) write all identified recipes.

But I think it would be easier to follow if we have separate stories for each page type, so I have now done that:

* [CSS types](https://github.com/mdn/stumptown-content/issues/441)
* [CSS selectors](https://github.com/mdn/stumptown-content/issues/440)
* [CSS functions](https://github.com/mdn/stumptown-content/issues/439)
* [CSS at-rules](https://github.com/mdn/stumptown-content/issues/466)
* [CSS at-rule descriptors](https://github.com/mdn/stumptown-content/issues/467)

That leaves this story to cover CSS properties only.

# Acceptance Criteria 
- [x] recipes for CSS property pages are written and checked into stumptown-content
- [x] spec for how to lint with these recipes is written and checked into stumptown-content

wbamberg:
In this comment I'll analyze which different types of pages we have in the CSS documentation.

We have 1001 pages under https://developer.mozilla.org/en-US/docs/Web/CSS, which you can find listed [here](https://wiki.developer.mozilla.org/en-US/docs/User:wbamberg/all-css-pages).

Because the CSS page hierarchy is mostly flat, and our tagging is a mess, it's not that simple to figure out which pages we have or what sorts of things they document.

Page titles seem to be the most reliable clue here. I've used the following rules:

* guides start with an uppercase letter
* types start with "<"
* at-rules start with "@"
* pseudoclasses and pseudoelements start with ":"
* properties and at-rule descriptors start with lowercase letter or "-", and don't end with "()"
* functions start with a lowercase letter and end with "()"

This gives us something like:
* 192 guides
* 135 pseudoclasses and pseudoelements
* 35 types
* 71 functions
* 13 at-rules

= 446, leaving 555 properties and at-rule descriptors. Of those I think we have:

* 420 non-prefixed properties
* 86 prefixed properties
* leaving 49 at-rule descriptors

This suggests we could have the following page types/recipes:

* `guides`
* `css-property`
* `css-selector` (including both pseudoclasses and pseudoelements)
* `css-type`
* `css-at-rule`
* `css-function`
* `css-at-rule-descriptor`

chrisdavidmills:
Looks pretty good, @wbamberg . I can't immediately think of any major page types you've missed.

The only outlier that immediately sprints to mind is the custom properties ref page:

[...]

PR #454, opened 2020-05-25
Add recipes for CSS properties

wbamberg:
This PR adds recipes for CSS properties, as discussed in https://github.com/mdn/stumptown-content/issues/396.

Apart from the substantive additions there's some rework, which I put into separate commits to (I hope) make it clearer.

Currently we have a spec called "javascript-linter-spec.md", but I wanted to have just one spec across all domains, since they share so many ingredients. But since we probably link to "javascript-linter-spec.md" from various spots, I wanted to keep that document around. So:

* https://github.com/mdn/stumptown-content/commit/2421ad415d192fe896d1be3158b7546dd633276d creates a new doc "linter-spec.md", starting out with the same content as "javascript-linter-spec.md"
* https://github.com/mdn/stumptown-content/commit/1c0553d933d638f4845d9f56ffdb4b4f3791830d updates "javascript-linter-spec.md" to be just a pointer to the new document
* https://github.com/mdn/stumptown-content/commit/16cecd8878d5c16f4fd8ba2aecc707635a50d499 makes "linter-spec.md" not be JS-specific
*  https://github.com/mdn/stumptown-content/commit/7c291cef72114941f262d91252f0dba18fb6685e alphabetises the ingredient descriptions in "linter-spec.md"
* https://github.com/mdn/stumptown-content/commit/d35b2328f938e486938be65671d451d7b8d04ce3 and https://github.com/mdn/stumptown-content/commit/6adbdfe686db29bb8f8aad7e47e00dd14e630462 actually add the new recipes and descriptions

There are two recipes, one for normal CSS properties and one for shorthand properties, and three new ingredients:

* `data.constituent_properties`
* `data.formal_definition`
* `data.formal_syntax`

## Constituent properties

`data.constituent_properties` is only for shorthand properties, and is only allowed to list the properties the shorthand contains.

## Formal definition and syntax

Normal properties get two extra ingredients:
* `data.formal_definition`: this contains the `{{CSSInfo}}` call
* `data.formal_syntax`: this contains the `{{CSSSyntax}}` call

After some discussion in https://github.com/mdn/stumptown-content/issues/396 and also in https://github.com/mdn/stumptown-content/issues/440#issuecomment-628365377, I'm proposing to make these things be separate H2 headings, rather than have one ingredient "Formal definition and syntax". This is because it seems plausible to expect that other CSS things might want formal syntax but not formal definition.

We could break the linkage between H2 headings and ingredients, and allow ingredients to live anywhere in the outline. But having thought a bit about this I think it's more trouble than it's worth. Suppose CSS properties wanted a structure like:

```
## Formal definition and syntax
    ### Formal definition
    ### Formal syntax
```

Then suppose CSS selectors wanted to keep formal syntax, but not formal definition. Then they might want it to live in a top-level H2:

```
## Formal syntax
```

Then the linter would need to apply different rules for the same ingredient, depending on the recipe it's living in. This breaks the rule that the definition of an ingredient is independent of the recipes that uses it, and this rule seems important for keeping things simple. It means that once you understand the definition, you can use it anywhere.


project-docs/linter-spec.md
```diff
@@ -0,0 +1,186 @@
...
+Each example may be a simple example or a live sample. If the example contains a call to the `{{EmbedLiveSample}}` macro, it is a live sample. Otherwise it is a simple example.
+
+Live samples must satisfy certain additional constraints:
+
+- The call to `{{EmbedLiveSample}}` must be the last element in the example (the only exception is that text nodes consisting only of newlines may follow the `{{EmbedLiveSample}}` call). This implies that there may only be one live sample in a given H3 section, and that the iframe showing the output must appear last in the rendered page, after any code or explanatory prose.
+- The ID parameter passed to `{{EmbedLiveSample}}` must match the ID of the example's H3 heading. This implies that the example is not allowed to contain any code blocks that are not included in the live sample.
+
+#### data.formal_definition
+
+To satisfy this ingredient a page must have a section demarcated by `H2#Formal_definition`.It must contain only a call to the `{{CSSInfo}}` macro.
```
ddbeck, line 155:
```suggestion
To satisfy this ingredient a page must have a section demarcated by `H2#Formal_definition`. It must contain only a call to the `{{CSSInfo}}` macro.
```

project-docs/linter-spec.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### data.formal_syntax
+
+To satisfy this ingredient a page must have a section demarcated by `H2#Formal_syntax`.It must contain only a call to the `{{CSSSyntax}}` macro.
```
ddbeck, line 159:
```suggestion
To satisfy this ingredient a page must have a section demarcated by `H2#Formal_syntax`. It must contain only a call to the `{{CSSSyntax}}` macro.
```

ddbeck (approved):
I noticed the two tiniest of formatting problems, but other than that, this looks really good. Thanks for putting in a lot of good work on this, Will.

(This was suppose to be "Changes requested" but I goofed. Please consider this approved pending the suggestions below.)

ddbeck (approved):

ddbeck:
I wasn't sure if @Elchi3 also wanted a look at this first, but this is good to merge, if not.

project-docs/linter-spec.md
```diff
@@ -0,0 +1,186 @@
...
+- [javascript-namespace](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-namespace.yaml)
+- [javascript-constructor](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-constructor.yaml)
+- [javascript-method](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-method.yaml)
+- [javascript-property](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-property.yaml)
+- [javascript-error](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-error.yaml)
+- [javascript-language-feature](https://github.com/mdn/stumptown-content/blob/master/recipes/javascript-language-feature.yaml)
+
+## Recipe ingredients
+
+In total these recipes use the following ingredients:
```
Elchi3, line 47: Above you only mention JS recipes but then here you list all ingredients for all recipes. 
Should we also list CSS recipes?

Elchi3 (commented):
Looks good to me, thanks Will! 
Just one comment that isn't blocking but that caught me while reading.

project-docs/linter-spec.md
wbamberg: Oops, that should be fixed now.

Elchi3 (approved):

Issue #447, opened 2020-05-20
Fix ordering linting errors in JS docs


Elchi3:
https://github.com/mdn/stumptown-content/pull/438 has landed and is quite brutal to us in terms of new errors in JS docs. It seems that over 300 errors are now about wrong ordering.

Elchi3:
From the call just now:

Errors are thrown quite verbosely (multiple per page), so this is as bad as I thought. Will is going to look into making it more silent.

Also, one thing to look into is Polyfill sections and whether we should add it as optional ingredient.

wbamberg:
I've filed https://github.com/mdn/stumptown-content/issues/448 to make the linter log only one error when sections are out of order.

wbamberg:
Does anyone want to make an argument for keeping this section: https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString#Firefox-specific_notes ?

wbamberg:
I also filed https://github.com/mdn/stumptown-content/issues/449 about making polyfill an ingredient.

Elchi3:
Thank you so much for digging into content linting, Will! I'm super happy about this discussion on content structures! All what you've found is great insight!

---

> Mostly this is relocating H2#Polyfill but there were a few other obvious changes too.

I see Polyfills are now above Examples. Not sure Polyfills are more important than Examples, but I don't know if our section order is by importance or what the criteria is... We probably need to dig into this some more at a later point anyway, so thanks for fixing this for the moment and let's discuss in the issue you've filed and see how we could address this properly.

---

> H2#Syntax for properties

I agree to everything you say about these. We agreed back when we first created JS recipes that syntax sections for properties make no sense. Also agree about the other cases and for String: thanks for taking the time to rewrite this!

---

> Old compat notes

I think we should apply the BCD criteria for "[irrelevant features](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features)" here as well. If these notes talk about things that date back more than two years ago (roughly older than Firefox 60 at this point), we should remove it.

---

> TypedArray

[...]

PR #461, opened 2020-05-27
Update the linter to accept classes with no constructor, if this is explicitly mentioned

wbamberg:
See https://github.com/mdn/stumptown-content/issues/447#issuecomment-632927388.

This makes the linter accept, as an alternative to a `<dl>` containing a link to  constructor, a `<p>` that starts with text like "This object cannot be instantiated directly." (which is expected to be followed by some more text providing extra context).

This will enable us to fix linter errors for:

* [TypedArray](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
* [Generator](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator)

### DIFF
--- a/project-docs/linter-spec.md
+++ b/project-docs/linter-spec.md
@@ -124,14 +124,16 @@ To satisfy this ingredient a page must have a section demarcated by `H2#Constitu
 
 #### data.constructor
 
-To satisfy this ingredient, a page must meet the following requirements:
+To satisfy this ingredient, a page must contain a section demarcated by `H2#Constructor` that contains one of the following:
 
-- A section demarcated by `H2#Constructor` that contains only a `<dl>` element.
-- The `<dl>` must contain a single `<dt>` followed by a single `<dd>`.
-- The `<dt>` must contain either:
-  - only a single `<code>` element, that contains only a single `<a>` element
-  - only a call to the `{{jsxref}}` macro.
-- The `<dd>` must contain `Creates a new <code>NameOfTheObject</code> object.` where `NameOfTheObject` is replaced with the actual name.
+1. Only a `<dl>` element.
+   - The `<dl>` must contain a single `<dt>` followed by a single `<dd>`.
+   - The `<dt>` must contain either:
+     - only a single `<code>` element, that contains only a single `<a>` element
+     - only a call to the `{{jsxref}}` macro.
+   - The `<dd>` must contain `Creates a new <code>NameOfTheObject</code> object.` where `NameOfTheObject` is replaced with the actual name.
+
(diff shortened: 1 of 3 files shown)


project-docs/linter-spec.md
```diff
@@ -124,14 +124,16 @@ To satisfy this ingredient a page must have a section demarcated by `H2#Constitu
...
-  - only a call to the `{{jsxref}}` macro.
-- The `<dd>` must contain `Creates a new <code>NameOfTheObject</code> object.` where `NameOfTheObject` is replaced with the actual name.
+1. Only a `<dl>` element.
+   - The `<dl>` must contain a single `<dt>` followed by a single `<dd>`.
+   - The `<dt>` must contain either:
+     - only a single `<code>` element, that contains only a single `<a>` element
+     - only a call to the `{{jsxref}}` macro.
+   - The `<dd>` must contain `Creates a new <code>NameOfTheObject</code> object.` where `NameOfTheObject` is replaced with the actual name.
+
+2) A `<p>` element that starts with the text "This object cannot be instantiated directly.".
```
ddbeck, line 136:
```suggestion
2. A `<p>` element that starts with the text "This object cannot be instantiated directly.".
```

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
```diff
@@ -26,11 +41,20 @@ function handleDataConstructor(tree, logger) {
     return null;
   }
 
+  const section = utils.sliceSection(heading, tree);
+
+  // Constructor sections are allowed to have no actual links
+  // to constructors, if they explicitly record this fact
+  let ok = checkNoConstructor(section.children);
```
ddbeck, line 48: A small style question here: Is the checkNoConstructor call significant after this line? I'm not sure why it shares a variable with the result of `checkLinkList`. Is there a reason you didn't go with `if (checkNoConstructor(…))` on the next line?

ddbeck (changes requested):
Looks good. One tiny correction and a style question, but no showstoppers here. 👍

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js
wbamberg: No, you're right, that is simpler and I've made the change.

ddbeck (approved):
Thank you, Will! 🎉

Issue #411, opened 2020-04-30
Fix linting errors "Expected h2#Examples for data.examples"


Elchi3:
We need to update the JS docs to have example sections. This affects quite some pages:

---
20 errors      javascript-method/data.examples/expected-heading                                            
17 errors      javascript-constructor/data.examples/expected-heading                                       
15 errors      javascript-language-feature/data.examples/expected-heading                                  
9 errors       javascript-class/data.examples/expected-heading                                             
8 errors       javascript-property/data.examples/expected-heading                                          
6 errors       javascript-namespace/data.examples/expected-heading                  

= 75

Elchi3:
Okay, I've done some good progress on this and brought the issue number down to 30 (from 75). The remaining work can be put in 3 categories:

### Single page describing multiple features
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_fields (we already said that this page needs splitting into two)
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Assignment_Operators
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

### Structural rewrite needed:
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

### Examples entirely missing
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/@@toPrimitive
- [ ] https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl
[...]

PR #464, opened 2020-05-29
Remove data.examples from javascript-namespace recipe

ddbeck:
This removes `data.examples` from the namespace recipe for JavaScript pages. From https://github.com/mdn/stumptown-content/issues/411#issuecomment-635921346. I don't know if we should do this.

Pros:
- Namespaces can't be instantiated, so namespace examples may duplicate individual method and property pages
- As best I can tell, namespace pages didn't have any examples until we started fixing linter errors. It seems poor form for the recipe to require them, if no such page ever had examples before

Cons:
- It might be nice to have examples that tie different pieces of an namespace together

### DIFF
--- a/recipes/javascript-namespace.yaml
+++ b/recipes/javascript-namespace.yaml
@@ -11,7 +11,6 @@ body:
   - data.static_properties?
   - data.static_methods?
   - prose.*
-  - data.examples
   - data.specifications
   - data.browser_compatibility
   - prose.see_also


wbamberg:
Namespaces are a bit of a mixed bag. Some namespaces are more like "collections of constructors" (e.g. [`Intl`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) and some are more like "classes with only static methods" (e.g. [`Relect`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect).

If we don't expect examples for "classes with static methods" if seems inconsistent for us to expect them for instantiable classes like `Array` and `String`.

And yes, there will probably be duplication of examples, but again the same applies to examples in class pages.

So unless it gets really ugly to add examples for these things I think it would be preferable to have them.

ddbeck:
Is there a model namespace example? Maybe the real problem is that I don't where to start when writing examples for a namespace.

wbamberg:
I haven't seen any that I like enough to call them a model, but the examples for [`Math`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math#Examples) are quite nice.

ddbeck:
OK, thanks!

Let's not do this. Closing this PR.

PR #474, opened 2020-06-08
Map page tags to CSS property recipes

wbamberg:
We don't have a complete set of CSS page recipes yet. But we do have recipes for CSS properties and CSS shorthand properties. So this PR adds mappings from page tags to these two recipes.

It uses:
* ["CSS", "Property"] to identify a CSS property page
* ["CSS", "Shorthand property"] to identify a CSS shorthand property page

Having this landed would unblock work to fix all the page tags for these types, which is a possible work item for a doc sprint in the virtual all-hands.

Having page tags fixed is also of course a prerequisite for being able to generate linter errors for these pages, and fixing those errors is another possible work item for a doc sprint.

It's worth considering whether the tags I've chosen here are correct. In particular this choice means we don't tag shorthand properties with "Property", which could potentially break macros that rely on tags. IIRC our CSS docs rely on tags more than other areas, because of the completely flat hierarchy. But from a quick look at macros like [CSSInfo](https://github.com/mdn/kumascript/blob/master/macros/CSSInfo.ejs), [CSSSyntax](https://github.com/mdn/kumascript/blob/master/macros/CSSSyntax.ejs), [CSSRef](https://github.com/mdn/kumascript/blob/master/macros/CSSRef.ejs), [CSS_Ref](https://github.com/mdn/kumascript/blob/master/macros/CSS_Ref.ejs), noone there seems to be relying on "Property" as a tag name. And our tagging is such a mess that anything that relies on correct tags is probably broken anyway.

As an alternative we could give shorthand properties something like ["CSS", "Property", "Shorthand"]. But then the linter would need to be cleverer in how it applies the mapping. Maybe it's worth it?

[In a separate commit](https://github.com/mdn/stumptown-content/commits?author=wbamberg), I also alphabetised the array of mappings, because otherwise things will get weird.

### DIFF
--- a/scripts/scraper-ng/plugins/identify-recipes.js
+++ b/scripts/scraper-ng/plugins/identify-recipes.js
@@ -55,6 +55,18 @@ const signatures = [
       tags: ["Guide"],
     },
   },
+  {
+    recipePath: recipe("css-shorthand-property"),
+    conditions: {
+      tags: ["CSS", "Shorthand property"],
+    },
+  },
+  {
+    recipePath: recipe("css-property"),
+    conditions: {
+      tags: ["CSS", "Property"],
+    },
+  },
 ];
 
 /**
--- a/scripts/scraper-ng/plugins/identify-recipes.js
+++ b/scripts/scraper-ng/plugins/identify-recipes.js
@@ -2,9 +2,21 @@ const path = require("path");
 
 const signatures = [
   {
-    recipePath: recipe("javascript-namespace"),
+    recipePath: recipe("css-property"),
     conditions: {
-      tags: ["JavaScript", "Namespace"],
+      tags: ["CSS", "Property"],
+    },
+  },
+  {
+    recipePath: recipe("css-shorthand-property"),
+    conditions: {
+      tags: ["CSS", "Shorthand property"],
+    },
+  },
+  {
+    recipePath: recipe("guide"),
+    conditions: {
+      tags: ["Guide"],
     },
   },
   {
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
These tags seem reasonable to me. Initially it seems weird to not have shorthand properties tagged with "Property", but then again I guess it doesn't matter that much, as long as the system works.

wbamberg:
> it seems weird to not have shorthand properties tagged with "Property"

It does seem weird. It is weird. One alternative would be to update the linter so it chooses the recipe with the most tags that match, so:

* given a page with ["CSS", "Property", "Shorthand property", "Something else"] it chooses "css-shorthand-property"
* given a page with ["CSS", "Property", "Something else"] it chooses "css-property"

Perhaps we should do that.

wbamberg:
> use a recipe-specific tag (e.g., "recipe:css-shorthand-property")

I vote to do this for the CSS pages, anyway, and maybe do it for every page. The only real argument against is that it's a lot of work, but i don't think it's *that* much work really. Just really boring work :).

chrisdavidmills:
> I vote to do this for the CSS pages, anyway, and maybe do it for every page. The only real argument against is that it's a lot of work, but i don't think it's that much work really. Just really boring work :).

And the kind of work that's appropriate for volunteers to help with. I vote for this too.

Elchi3:
Agreed, and as Daniel said, this should be very nice for explicit recipe declaration in future markdown front matters. 
It is a good signal to all writers and contributors: We can say: This page is of type "recipe:css-shorthand-property and thus the following authoring rules apply: [...]

wbamberg:
OK, I've made this change.

ddbeck (approved):
I'm happy with this. Marking approved, but I'll leave the PR open in case Chris and Florian have anything to add.

Elchi3 (approved):
Let's do this :tada:

PR #477, opened 2020-06-09
Add handler for data.formal_syntax

### LINKED ISSUE #458: U - Write ingredient handler for `data.formal_syntax`
From #454, we'll need an ingredient handler for `data.formal_syntax`.

## Acceptance criteria

- [ ] the linter lints CSS pages for the `data.formal_syntax` ingredient
- [ ] the ingredient handler conforms to the specification

ddbeck:
This PR adds an ingredient handler for `data.formal_syntax`. It also amends the spec for the handler slightly, to accommodate the obligatory `<pre class="syntaxbox">` tag required for the `CSSSyntax` macro (see https://github.com/mdn/stumptown-content/issues/458#issuecomment-639716348).

I also sorted the ingredient handlers in `ingredient-handlers/index.js` while I was here.

Fixes #458.

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-syntax.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-syntax.js
@@ -1,3 +1,77 @@
-function handleDataFormalSyntax(tree, logger) {}
+const { isMacro, sliceSection } = require("./utils");
+const { select } = require("hast-util-select");
+const visit = require("unist-util-visit");
+
+function handleDataFormalSyntax(tree, logger) {
+  const id = "Formal_syntax";
+  const body = select(`body`, tree);
+  const heading = select(`h2#${id}`, body);
+  const section = sliceSection(heading, body);
+
+  if (heading === null) {
+    logger.expected(body, `h2#${id}`, "expected-heading");
+    return null;
+  }
+
+  // The first non-whitespace node after the `h2` must be pre.syntaxbox
+  const relevantChildren = section.children.filter(
+    (child) => child.tagName !== "h2" && !isWhiteSpaceTextNode(child)
+  );
+  const expectedPre = relevantChildren.length > 0 ? relevantChildren[0] : null;
+  if (
+    expectedPre === null ||
+    expectedPre.type !== "element" ||
+    expectedPre.tagName !== "pre" ||
+    !Array.isArray(expectedPre.properties.className) ||
+    !expectedPre.properties.className.includes("syntaxbox")
+  ) {
+    logger.expected(tree, expectedPre, "expected-pre.syntaxbox");
+    return null;
+  }
+
+  // The `pre` tag's first child must be a macro node
+  const expectedMacro =
+    expectedPre.children.length > 0 ? expectedPre.children[0] : null;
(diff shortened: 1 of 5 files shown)


scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-syntax.js
```diff
@@ -0,0 +1,79 @@
...
+      "No other elements allowed in formal syntax",
+      "syntax-only"
+    );
+    return null;
+  }
+
+  return heading;
+}
+
+function isWhiteSpaceTextNode(node) {
```
wbamberg, line 75: Consider putting this in utils.js?

scripts/scraper-ng/test/ingredient-data.formal_syntax.test.js
```diff
@@ -0,0 +1,73 @@
...
+    const file = process(valid, recipe);
+
+    expect(file).not.hasMessageWithId(ingredientName);
+    expectPositionElement(file.data.ingredients[0], ingredientName, "h2");
+  });
+
+  test("missing h2", () => {
+    const file = process("", recipe);
+
+    expect(file.messages.length).toBeGreaterThan(0);
```
wbamberg, line 23: I don't really understand the purpose of these `toBeGreaterThan` checks. Here (and in any other error case where the argument is 0) it's redundant, since we immediately afterwards check for a particular message.

scripts/scraper-ng/test/ingredient-data.formal_syntax.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+    expectNullPosition(file.data.ingredients[0], ingredientName);
+  });
+
+  test("missing syntaxbox class", () => {
+    const missingClass = `<h2 id="Formal_syntax">Formal syntax</h2>
+                          <pre>{{ csssyntax("display") }}</pre>`;
+    const file = process(missingClass, recipe);
+
+    expect(file.messages.length).toBeGreaterThan(1);
```
wbamberg, line 45: Here we check `toBeGreaterThan(1)`. So because there's an error and a CSSSyntax macro call, then there are two messages. But in the previous test ("missing pre"), the same applies but we only check `toBeGreaterThan(0)`. I don't understand why.

I also don't understand, if we're going to check the number of messages, why not check the exact number rather than the weaker "greater than". But in general, I don't think it's worth checking the number of errors here at all.

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-syntax.js
```diff
@@ -0,0 +1,79 @@
...
+const { select } = require("hast-util-select");
+const visit = require("unist-util-visit");
+
+function handleDataFormalSyntax(tree, logger) {
+  const id = "Formal_syntax";
+  const body = select(`body`, tree);
+  const heading = select(`h2#${id}`, body);
+  const section = sliceSection(heading, body);
+
+  if (heading === null) {
```
wbamberg, line 11: It's odd to me to have this check *after* we try to use `heading` at line 9.

wbamberg (changes requested):
Hey @ddbeck , this looks great. I had some really minor gripes.

wbamberg (approved):
Thanks Daniel, looks great to me!

PR #480, opened 2020-06-16
Add data.formal_definition handler

### LINKED ISSUE #457: U - Write ingredient handler for `data.formal_definition`
From #454, we'll need an ingredient handler for `data.formal_definition`.

## Acceptance criteria

- [x] the linter lints CSS pages for the `data.formal_definition` ingredient
- [ ] the ingredient handler conforms to the specification

ddbeck:
This PR adds a complete ingredient handler for `data.formal_definition`. It also slightly amends the spec for the ingredient, to cover the `<p>` tag which Kuma wraps around a bare `{{ CSSInfo }}` call.

Fixes #457.

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-definition.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-definition.js
@@ -1,0 +1,7 @@
+function handleDataFormalDefinition(tree, logger) {
+  const id = "Formal_definition";
+  const body = select(`body`, tree);
+  const heading = select(`h2#${id}`, body);
+}
+
+module.exports = handleDataFormalDefinition;
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-definition.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-definition.js
@@ -1,48 +1,49 @@
 const { findUnexpectedNode, isMacro, sectionHandler } = require("./utils");
 
-const handleDataFormalDefinition = sectionHandler("Formal_definition", handle);
+const handleDataFormalDefinition = sectionHandler(
+  "Formal_definition",
+  (section, logger) => {
+    const heading = section.children[0];
 
-function handle(section, logger) {
-  const heading = section.children[0];
-
-  // Find the first P with a CSSInfo macro as one of its children
-  let expectedMacro;
-  let expectedP = section.children.find((node) => {
-    if (node.tagName === "p" && node.children) {
-      // Get first CSSInfo child node, if it exists
-      expectedMacro = node.children.find((node) => isMacro(node, "CSSInfo"));
-      if (expectedMacro) {
-        return true;
+    // Find the first P with a CSSInfo macro as one of its children
+    let expectedMacro;
+    let expectedP = section.children.find((node) => {
+      if (node.tagName === "p" && node.children) {
+        // Get first CSSInfo child node, if it exists
+        expectedMacro = node.children.find((node) => isMacro(node, "CSSInfo"));
+        if (expectedMacro) {
+          return true;
+        }
       }
+      return false;
+    });
+
+    if (!expectedP) {
+      logger.expected(
+        section,
+        "CSSInfo macro paragraph",
+        "expected-cssinfo-macro"
+      );
+      return null;
     }
-    return false;
-  });
 
-  if (!expectedP) {
-    logger.expected(
+    // The section must contain only a heading and `<p>{{CSSInfo}}</p>`
+    const extraneousNode = findUnexpectedNode(
       section,
-      "CSSInfo macro paragraph",
-      "expected-cssinfo-macro"
+      [heading],
+      [expectedP, expectedMacro]
     );
-    return null;
-  }
+    if (extraneousNode !== null) {
+      logger.fail(
+        extraneousNode,
+        "No other elements allowed in data.formal_definition",
+        "expected-macro-only"
+      );
+      return null;
+    }
 
-  // The section must contain only a heading and `<p>{{CSSInfo}}</p>`
-  const extraneousNode = findUnexpectedNode(
-    section,
-    [heading],
-    [expectedP, expectedMacro]
-  );
-  if (extraneousNode !== null) {
-    logger.fail(
-      extraneousNode,
-      "No other elements allowed in data.formal_definition",
-      "expected-macro-only"
-    );
-    return null;
+    return heading;
   }
-
-  return heading;
-}
+);
 
 module.exports = handleDataFormalDefinition;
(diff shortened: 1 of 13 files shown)


ddbeck:
Pushed a change that makes a utility out of finding extraneous nodes in a section. It's used here, in `data.formal_syntax`, and soon in `data.constituent_properties`.

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/utils.js
```diff
@@ -49,6 +49,39 @@ function sliceBetween(startNode, endCondition, tree) {
...
+ * @param {Object} tree - the tree to search for unexpected nodes
+ * @param {Array} expectedTrees - An array of nodes such that each node itself
+ *   and all of its descendants are permited. An expected tree's descendants are
+ *   never visited.
+ * @param {Array} expectedNodes - An array of nodes such that each node is
+ *   permitted. An expected node's descendants are visited, if they exist. The
+ *   starting node and white space are always expected.
+ * @returns {Object|null} a node or `null`
+ */
+function findExtraneousNode(tree, expectedTrees, expectedNodes) {
```
wbamberg, line 64: This is really nitpicky but "Unexpected" might be a clearer word than "Extraneous".

ddbeck: ✅ Renamed `findExtraneousNode` to `findUnexpectedNode`

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-definition.js
```diff
@@ -0,0 +1,58 @@
...
+  const body = select(`body`, tree);
+  const heading = select(`h2#${id}`, body);
+
+  if (heading === null) {
+    logger.expected(body, `h2#${id}`, "expected-heading");
+    return null;
+  }
+
+  // Section must have an immediate child P element containing the CSSInfo macro
+  const section = sliceSection(heading, body);
```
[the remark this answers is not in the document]
ddbeck: Yeah, this is the right thing to do. Though to do it kinda gracefully ended up being a slightly bigger change than I anticipated. It might help to look at that commit alone. If you want me to move that change to its own PR, let me know.

wbamberg:
Thanks for this refactoring, @ddbeck ! All this code seems to work fine.

The pattern here implements the common code as a kind of wrapper round the ingredient-specific code, so we do:

```js

const handleXYZ = utils.requiredSectionHandler(XYZ_id, handle);

function handle(tree, logger, section, handler) {
// ingredient-specific handling
}
```

This seems a bit harder to follow than just implementing is as a separate utility function that returns a section or null:

```js
// in utils:
function requiredSectionHandler(id, tree, logger) {
    // Extract the section and heading
    const body = select("body", tree);

    const heading = select(`h2#${id}`, body);
    if (heading === null) {
      logger.expected(body, `h2#${id}`, "expected-heading");
      return null;
    }

    return sliceSection(heading, body);
}

// in the XYZ ingredient handler
function handleXYZ() {
  const section = requiredSectionHandler(XYZ_id, tree, logger);
  if (section === null) {
    return section;
  }
//  ingredient-specific handling
```

Having the second pattern means the utility code is more loosely-coupled with the code that uses it, as well as being (I think) easier to follow. It means for example that you have to pass through all arguments which any handler might need, even when most of them don't, and this means more state gets spread throughout the code.

But perhaps I'm missing something here.

Even then though I'm not sure we need to pass `tree` into the handler. I looked at this and we call `requiredSectionHandler` from 6 ingredient handlers.

* In four of them - data-browser-compatibility, data-browser-compatibility, data-formal-definition, and data-specifications - it's not used at all.

* In one of them - data-formal-syntax - it's used to identify the node in a logger.expected call: `logger.expected(tree, expectedSyntaxBox, "expected-macro");` - but I think this should use `section` instead.

* The other case  - data-constructor - is a weird one. It passes `tree` into `checkLinkList`, [which uses it to fetch the `section`](https://github.com/mdn/stumptown-content/blob/master/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/link-list-checker.js#L21-L30) - again! The difference being that here we don't require the section, so we don't log an error if it's not found. So I wonder if it would be better to make `requiredSectionHandler` be just `sectionHandler` and give it a `isRequired` boolean flag, then we could use that in data-class-members as well?

Anyway. Like I say, this code works fine as it is so I don't know how much these comments matter. But if you wanted to consider them, it's up to you whether to do so in this PR or in a follow-up.

ddbeck:
I really appreciate this feedback, Will. I think I can demonstrate an improvement here, but I worry we're getting a little far afield, so I'm going to rework _only_ the `data.formal_definition` handler, then revisit the other handlers in a follow-up PR. I'd appreciate one more look here, but some discussion on my thinking follows.

You raised some important points. The section handler API is a bit sloppy. It adds some superfluous state (stuffing more arguments into the handler is probably a mistake). For example, if I'm writing an ingredient handler, I shouldn't really _need_ `tree` once I've got `section`. A better version of this is probably something like `sectionHandler(sectionId, handler, optional)`, and then `function handle(section, logger) { … }`. This isn't perfect (it doesn't have an answer for `heading`), but it's much tidier.

But to do the less sloppy API requires a more pervasive refactoring, changing lots of `logger` calls. It's a lot to follow, which is why I don't want to do it all at once.

---
An aside here: I wish JS had a Python-like decorator syntax. In Python, this would be much prettier as:

```python
@section_handler("Some_Id")
def some_ingredient_handler(section, logger):
    …
```

I guess this will look OK as an arrow function in JS, but it could be better!

```js
const someIngredientHandler = sectionHandler("Some_Id", (section, logger) => { … });
```
---

Finally, I started out with a strict utility function like you suggest as a second pattern. I started with something like this:

```js
function handleXYZ() {
  const {section, heading} = findSection("Section_Id", tree, logger);
  if (section === null) {
    return section;
  }
```

I found this tough to keep in my head. There's a lot to keep up with here:

- I want both `section` and `heading` in multiple places (for positioning messages, though this is suboptimal, as you mention), but it's sort of surprising to get both of those objects back from this utility. Almost all unist utils return a single node or an array of nodes, not some structured thing.
- It's also an oddity that `findSection()` has recording an error message as a possible side effect; I have to remember to _not_ log a message when using the results from `findSection`.

Plus, it doesn't spare me from having to do a check and return null, which is also part of the repetitive code. It felt a bit fussy to make so many changes, but not actually resolve the complete duplication.

wbamberg:
> A better version of this is probably something like sectionHandler(sectionId, handler, optional), and then function handle(section, logger) { … }. This isn't perfect (it doesn't have an answer for heading), but it's much tidier.

Yes, this is much tidier, and I agree with doing it in a follow-up.

We need `heading` to return from the handler, so we can do the ingredient order check. And I had at first thought that this means we have to pass it into the section handler. But maybe this could also be done in this wrapper/decorator. So the section handler could return ok/not ok, and the wrapper could return heading/null based on that, and that would save us from having to do things like https://github.com/mdn/stumptown-content/blob/master/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-constructor.js#L119-L123 in the section handler?

> I found this tough to keep in my head.

Fair enough.

ddbeck:
> section handler could return ok/not ok, and the wrapper could return heading/null based on that, and that would save us from having to do things like

Ah, yeah! Just return _something_ truthy or falsy from the handler and the wrapper can return the heading or `null`. Good idea. I'll include that in the refactor PR. But first, `data.constituent_properties`, coming shortly.

PR #481, opened 2020-06-16
Made data.formal_syntax messages human readable

ddbeck:
In #477, which implemented data.formal_syntax, I called `logger.expected` wrong (I passed in nodes, not strings, for the human-readable messages). This fixes those calls.

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-syntax.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-syntax.js
@@ -16,7 +16,7 @@ function handleDataFormalSyntax(tree, logger) {
   const section = sliceSection(heading, body);
   const expectedSyntaxBox = select("pre.syntaxbox", section);
   if (expectedSyntaxBox === null) {
-    logger.expected(tree, section, "expected-pre.syntaxbox");
+    logger.expected(tree, "pre.syntaxbox", "expected-pre.syntaxbox");
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-formal-syntax.js
```diff
@@ -16,7 +16,7 @@ function handleDataFormalSyntax(tree, logger) {
   const section = sliceSection(heading, body);
   const expectedSyntaxBox = select("pre.syntaxbox", section);
   if (expectedSyntaxBox === null) {
-    logger.expected(tree, section, "expected-pre.syntaxbox");
+    logger.expected(tree, "pre.syntaxbox", "expected-pre.syntaxbox");
```
wbamberg, line 19: I guess it would be possible to give a more useful node than `tree`? We're not terribly consistent about this, but for example for the `data.specifications` handler we pass in the H2: https://github.com/mdn/stumptown-content/blob/master/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-specifications.js#L33.

ddbeck: Good catch. I pushed a change to improve both of these `tree` cases in formal_syntax. I'll see if I can't clean up more of these (in other ingredients) in a follow up PR

wbamberg (approved):
Thanks Daniel! This looks good to me.

PR #482, opened 2020-06-16
Use correct IDs for placeholder formal syntax and constituent properties

ddbeck:
In #478, I accidentally used spaces instead of underscores in the IDs for `h2#Formal_syntax` and `h2#Constituent_properties`. This is a small fix ahead of #480 and the eventual PR for `data.constituent_properties`.


wbamberg (approved):
Thanks Daniel!

PR #488, opened 2020-06-23
Add recipe for CSS selectors

wbamberg:
This PR adds a recipe for CSS selectors, following the discussion in https://github.com/mdn/stumptown-content/issues/440.

I've omitted `data.formal_syntax` but included an optional `prose.applicable_properties`. This section is currently only included in the following pages:

https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::cue
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::cue-region
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::first-letter
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::first-line
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::grammar-error
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::marker
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::selection
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::spelling-error

(where it's named "Allowed properties" but I like "Applicable properties" better)

...out of about 70 non-prefixed selectors. However, it seems like a legitimately consistent and useful thing to want to include, and in future we might perhaps want to add it to more pages. We should probably structure it as a `data.` ingredient in future, as a list of property names. Or maybe we should do that now?

I also used generic tags `["CSS", "Selector"]` instead of a custom tag like `["recipe:css-selector"]` as we have for properties. I'm not sure if that's a good idea or not. But I'm still a little uncomfortable with these custom tags for some reason, and the reasoning that pushed us there with properties doesn't seem to apply here.

### DIFF
--- a/recipes/css-selector.yaml
+++ b/recipes/css-selector.yaml
@@ -6,7 +6,7 @@ body:
   - prose.short_description
   - prose.syntax
   - prose.description?
-  - prose.allowable_properties?
+  - prose.applicable_properties?
   - prose.accessibility_concerns?
   - prose.*
   - data.examples
--- a/recipes/css-selector.yaml
+++ b/recipes/css-selector.yaml
@@ -1,0 +1,15 @@
+data:
+  - title
+  - short_title?
+  - mdn_url
+body:
+  - prose.short_description
+  - prose.syntax
+  - prose.description?
+  - prose.allowable_properties?
+  - prose.accessibility_concerns?
+  - prose.*
+  - data.examples
+  - data.specifications
+  - data.browser_compatibility
+  - prose.see_also
--- a/recipes/css-selector.yaml
+++ b/recipes/css-selector.yaml
@@ -6,7 +6,7 @@ body:
   - prose.short_description
   - prose.syntax
   - prose.description?
-  - prose.applicable_properties?
+  - data.permitted_properties?
   - prose.accessibility_concerns?
   - prose.*
   - data.examples
--- a/project-docs/linter-spec.md
+++ b/project-docs/linter-spec.md
@@ -66,6 +66,7 @@ prose ingredients:
 data ingredients:
 
 - data.browser_compatibility
+- data.permitted_properties?
 - data.constituent_properties
 - data.constructor
 - data.constructor_properties?
@@ -115,6 +116,16 @@ Data ingredients typically impose more detailed requirements on pages than prose
 #### data.browser_compatibility
 
 To satisfy this ingredient a page must have a section demarcated by `H2#Browser_compatibility`. It must contain a call to the `{{Compat}}` macro.
+
+#### data.permitted_properties?
+
+This is an optional ingredient. It is present if the page has a section demarcated by `H2#Permitted_properties`. The section must contain only the following elements:
+
+1. A `<p>` element consisting of the text:
+
+   > Rules whose selectors include this element may only use the following CSS properties:
+
+2. An unordered list (a `<ul>` element) of two or more CSS property names. Each property name must be in the form `<li><a><code>p</code></a></li>`, where _p_ is the name of a property. The `<li>` elements must be in alphabetical order.
 
 #### data.constituent_properties
 
--- a/scripts/scraper-ng/plugins/identify-recipes.js
+++ b/scripts/scraper-ng/plugins/identify-recipes.js
@@ -8,6 +8,12 @@ const signatures = [
     },
   },
   {
+    recipePath: recipe("css-selector"),
+    conditions: {
+      tags: ["CSS", "Selector"],
+    },
+  },
+  {
     recipePath: recipe("css-shorthand-property"),
     conditions: {
       tags: ["recipe:css-shorthand-property"],


recipes/css-selector.yaml
```diff
@@ -0,0 +1,15 @@
+data:
+  - title
+  - short_title?
+  - mdn_url
+body:
+  - prose.short_description
+  - prose.syntax
```
ddbeck, line 7: It occurs to me now that `prose.syntax` could be more structured too (`data.syntax` could be a section + a code block, rather than arbitrary prose)

recipes/css-selector.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - prose.description?
+  - prose.applicable_properties?
```
ddbeck, line 9: I think we should go ahead and make this `data.applicable_properties`. It'll be very similar to `data.constituent_properties` so we may as well do it

ddbeck (commented):
This looks fine to me, but I am in favor of structuring parts of this (`data.applicable_properties` seems like a really good idea). I spotted another opportunity here too (with `prose.syntax`).

recipes/css-selector.yaml
wbamberg: This is a can of worms. For the JS docs we could also have chosen to structure `prose.syntax`. In method pages this might things like parameters and return value as well as a code block (see e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Syntax). In 

We decided that although we might want to add this structure in the future, for the time being we should just treat it as unstructured prose.

I think I'd prefer us to be consistent in this decision. Partly this is because the details of the structure might be different from one page type to another (for example, CSS selectors might just have code block+prose, while JS methods might have code block+parameters+returns). And it'll be easier to sort all this our when we have a complete set of `prose.syntax` examples to consider.

Also there's naming the ingredient. If the syntax structure for JS methods is different from that for CSS selectors then that's two different ingredients, which should have two different names. So we shouldn't call this `data.syntax`. We could perhaps have two syntax ingredients, `data.method_syntax` that includes parameters and return values, and `data.simple_syntax` or something better, that's just a code block and some optional prose. But this seems like the kind of conversation that's easier to have when we're looking at all the page types (including Web API and HTTP and the rest).

recipes/css-selector.yaml
wbamberg: https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/::first-letter#Allowable_properties, for example, presents properties in groups - font related, background-related, and so on.

If we specify it like `data.constituent_properties` then it's a list of links, 50-odd items, and no grouping. Is this OK or is it a significant loss? Also that page wants to annotate some links:

> `vertical-align` (only if float is none)

...so we'd either have to support that or lose this information.

recipes/css-selector.yaml
ddbeck: OK, let's not open this particular can just yet. I agree that it makes sense to put this off until we've looked at other page types.

recipes/css-selector.yaml
ddbeck: The more correct thing here would be to do some sort of user research to find out how developers use this information. Are they using it as a troubleshooting tool after the fact ("Why doesn't this property work? Ah, OK, it's in this list…") or are they reading it first? That'd probably tell us if groups or a list is more useful.

But going only on instinct here, I'd say that I don't much care for the grouped style. As a writer, I think I'd prefer some semi-free-form intro text that says something like, "Only a subset of CSS properties work with this pseudo-element. Supported properties include font, background, margin, padding, border, and text decoration properties" followed by an alphabetically sorted list of properties. That feels way more maintainable than trying to figure out which group (or which miscellaneous group, as in the `::first-letter` example) a property goes into.

As a reader, I think I'd prefer the list too. I don't trust myself to know what properties are in what groups, the property names are more widely recognizable than group descriptions, and I can skim a list faster than the current format.

For the annotations, I'm OK with losing them:

* The cue related pseudo-classes have some interesting behavior to do with borders and forming boxes. That seems notable enough for its own `prose.*` section.
* `::first-letter`'s float annotation is a reminder of a general rule about `vertical-align`: it only applies to inline-level elements. If `float` is not none, then it's not inline. The other one is "`word-spacing` (when appropriate)", which seems pretty pointless.
* `::selection` has a call out box for not supporting `background-image`. Assuming that was put in because it surprises people, that's a little loss, but I don't think we need to handle annotations generally for this one entry.

wbamberg: Thanks for this thoughtful response!

> The more correct thing here would be to do some sort of user research to find out how developers use this information. Are they using it as a troubleshooting tool after the fact ("Why doesn't this property work? Ah, OK, it's in this list…") or are they reading it first? That'd probably tell us if groups or a list is more useful.

In the absence of this my uninformed speculation is option 1, just because noone reads documentation until something doesn't work. But I definitely take your point that a list is easier to scan and doesn't make the assumption that our categorization of things matches the reader's.

wbamberg:
In the course of writing the description for this I thought `data.permitted_properties?` was a better name than `data.applicable_properties?`. I took the description mostly from `data.constituent_properties`.

ddbeck (approved):
Excellent! I really like the new name of the new ingredient too. 🎉

PR #490, opened 2020-06-23
Add handler for data.constituent_properties

### LINKED ISSUE #427: U - Write ingredient handler for `data.constituent_properties`
According to https://github.com/mdn/stumptown-content/issues/396 it look likely that we'll have a `data.constituent_properties` ingredient for CSS shorthand properties. We'll need to be able to lint for this ingredient.

## Acceptance criteria

* the linter lints CSS pages for the `data.constituent_properties` ingredient

ddbeck:
This PR implements the `data.constituent_properties` ingredient. It expects a section with fixed intro text, followed by an unordered list of two or more `<code>`-formatted links.

Fixes #427

### DIFF
--- a/project-docs/linter-spec.md
+++ b/project-docs/linter-spec.md
@@ -117,10 +117,13 @@ To satisfy this ingredient a page must have a section demarcated by `H2#Browser_
 
 #### data.constituent_properties
 
-To satisfy this ingredient a page must have a section demarcated by `H2#Constituent_properties`. This section must contain the following elements, in the order given below:
+To satisfy this ingredient a page must have a section demarcated by `H2#Constituent_properties`. The section must contain only the following elements:
 
-- a `<p>` element containing the following text: "This property is a shorthand for the following CSS properties:"
-- a `<ul>` element containing two or more `<li>` elements, each of which contains a single `<a><code>` element.
+1. A `<p>` element consisting of the text:
+
+   > This property is a shorthand for the following CSS properties:
+
+2. An unordered list (a `<ul>` element) of two or more CSS property names. Each property name must be in the form `<li><a><code>p</code></a></li>`, where _p_ is the name of a property. The `<li>` elements must be in alphabetical order.
 
 #### data.constructor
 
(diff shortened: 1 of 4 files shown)


wbamberg:
@ddbeck , I really like this. I think the `findUnexpectedNode` really pays off here, and makes the code very clear and solid.

But! I realise that we should also require that properties are listed in alphabetical order, just because we can and because it's a clear improvement. My fault for not including this in the spec!

By the way I also made a shorthand property page that passes the linting tests, so we can see what it looks like: https://developer.mozilla.org/en-US/docs/User:wbamberg/shorthand-property @Elchi3 , @chrisdavidmills . I think it is quite nice!

(Also, picking up that [conversation about making "See also" mandatory](https://github.com/mdn/stumptown-content/pull/493#issuecomment-648931207). The [`border`](https://developer.mozilla.org/en-US/docs/Web/CSS/border) page doesn't have "See also", but for my test page I added a link to the Learn content for borders. I think the ability to add cross-links like this is a real benefit of "See also", and encouraging us to use it more is a good thing. But that also makes me wonder if at some point we could explicitly structure this, giving reference pages an ingredient like "Learn area links" so we could render cross-links consistently...)

chrisdavidmills:
>  I think it is quite nice!

Agreed, this works pretty well. One thing that springs to mind is that there are a few cases (such as animation, and background) where the order of the constituents matters. Should we include this information somewhere, and if so, where?

ddbeck:
> But! I realise that we should also require that properties are listed in alphabetical order, just because we can and because it's a clear improvement. My fault for not including this in the spec!

This is a really good idea. I'll push a change to the spec and linter shortly.

> Agreed, this works pretty well. One thing that springs to mind is that there are a few cases (such as animation, and background) where the order of the constituents matters. Should we include this information somewhere, and if so, where?

I'm inclined to let other parts of the page cover this, in particular `prose.syntax` or `data.formal_syntax`. I looked at `animation` and `background` and I think there's a risk that an ordered list may imply a strict order that's not really there. For example, `animation` uses the _relative_ order of certain value types to set the longhand properties (e.g., the first duration is the animation duration, the second is the animation delay), but the animation name can appear in any position (so long as it doesn't conflict with a valid keyword value). There's enough complexity there that I don't think any one list could capture this aspect of shorthands.

wbamberg (approved):
Looks awesome, thank you Daniel!

Issue #441, opened 2020-05-14
U - Write recipe for CSS types


wbamberg:
This issue is split out of https://github.com/mdn/stumptown-content/issues/396, which wants us to write recipes for all the CSS reference documentation page types.

I thought to keep things a bit easier to follow it might make sense to have stories to track different CSS recipes. This one is for **CSS types**. 

# Acceptance Criteria 
- [ ] recipe for CSS types pages is written and checked into stumptown-content
- [ ] spec for how to lint with this recipe is written and checked into stumptown-content

wbamberg:
# CSS types

I count 35 pages for CSS types.

They have a form more or less like:

Short description
Interactive example (optional)
H2 Syntax
H2 Accessibility considerations (optional)
H2 Examples
H2 Specifications
H2 Browser compatibility
H2 See also

## Syntax

The Syntax section does not use `{{csssyntax}}` in any pages. It does use formal syntax-style a little bit. In general though, this section is a free-for-all, with lots of different structures.

Some types list items in Syntax that should perhaps be broken out into separate function pages (e.g. [`basic-shape`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/basic-shape) and [`easing-function`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/easing-function)).

It would be nice to design something consistent here, but so far we've treated H2#Syntax as unstructured prose. That looks like a better decision for JS methods, which are usually consistently structured anyway, than it does for pages like these. But I'm still not sure it's worth rewriting all these sections at this point - that feels like another project.

One interesting thing: the [`blend-mode`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/blend-mode) page uses a series of live samples inside Syntax. I think we ought to have an interactive example that covers some of the choices, and live samples in the (currently nonexistent) "Examples" section. But this made me think: if we are still interested in stumptown-compatibility, then our linter ought to complain about things like `{{EmbedLiveSample}}` calls occurring anywhere except the "Examples" section.

## Browser compatibility

The `display-*` types (e.g. [`display-inside`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/display-inside#Browser_compatibility)) have lots of BCD tables.

The [`custom-ident`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/custom-ident#Browser_compatibility) page has a BCD section but no table, and the following text:

[...]

Issue #439, opened 2020-05-13
U - Write recipe for CSS functions


wbamberg:
This issue is split out of https://github.com/mdn/stumptown-content/issues/396, which wants us to write recipes for all the CSS reference documentation page types.

I thought to keep things a bit easier to follow it might make sense to have issues to track different CSS recipes. This one is for **CSS functions**. 

# Acceptance Criteria 
- [ ] recipe for CSS function pages is written and checked into stumptown-content
- [ ] spec for how to lint with this recipe is written and checked into stumptown-content

wbamberg:
# CSS functions

I count 57 CSS function pages. They have more or less a form like:

Short description
Interactive example
H2. Syntax
H2. Accessibility concerns (occasionally)
H2. Examples
H2. Specifications
H2. Browser compatibility
H2. See also

## filter-function and transform-function

We can split the CSS functions into three groups:

* 11 under https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/filter-function/, that represent possible values for the `<filter-function>` type.
* 21 under https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/transform-function/, that represent possible values for the `<transform-function>` type
* the rest, which live directly under https://wiki.developer.mozilla.org/en-US/docs/Web/CSS

(Note that one function - `url()` - lived in two groups: [`CSS/url()`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/url())  and [`CSS/filter-function/url`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/filter-function/url). @a2sheppy just made the second of these redirect to the first one, which I think is the right call, but more filter-specific content should be added to the remaining page. I've filed https://github.com/mdn/sprints/issues/3243 for this.)

The pages under "CSS/filter-function/" are much thinner than the others, all lacking "Specifications" and "Browser compatibility".

The pages under "CSS/transform-function" include "Specifications" and "Browser compatibility" sections, but "Browser compatibility" just points to the [`<transform-function>`](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/transform-function#Browser_compatibility) page: the individual transform functions don't have their own compat data.

So we should decide whether the filter-function and transform-function pages should be treated differently from the other function pages, or not. If they should be treated the same, we should consider whether we should add BCD for them and any other missing bits. If they should be treated differently, what are they anyway?

I think we should treat them the same, and take on the work of giving them consistent structure and content.

[...]

PR #497, opened 2020-06-27
Add recipe for CSS functions

wbamberg:
Fix for https://github.com/mdn/stumptown-content/issues/439.

This PR adds a recipe for CSS functions. I think the issues for CSS functions are the same as those for CSS types:

1) do we want to standardise/structure "Syntax" at this time?
2) should these types of page contain BCD at all?
3) if not, should we deal with removing BCD from these pages at the moment?

Given that for CSS types we answered ["no, no, no"](https://github.com/mdn/stumptown-content/issues/441) I think we should have the same answer for CSS functions, so this PR adds a recipe that's just the same. In particular:

* `prose.syntax` is an unstructured prose block, as for all the other page types
* `data.browser_compatibility` is included, even though we would prefer eventually not to include BCD in pages like this.

### DIFF
--- a/recipes/css-function.yaml
+++ b/recipes/css-function.yaml
@@ -1,0 +1,15 @@
+data:
+  - title
+  - short_title?
+  - mdn_url
+body:
+  - prose.short_description
+  - data.interactive_example?
+  - prose.syntax
+  - prose.description?
+  - prose.accessibility_concerns?
+  - prose.*
+  - data.examples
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
Looks good. Thank you!

Elchi3 (approved):
Looks good to me, too. Consistent "no, no, no" makes sense :)

PR #499, opened 2020-06-30
Refactor ingredient handlers for sections

ddbeck:
This PR completes a refactor of the ingredient handlers which consume specific H2 sections from a page.

## What this does

Previously, we added a wrapper for ingredient handlers with a signature like this: `sectionHandler(sectionId, handlerFn, optional)`. The inner `handlerFn` is written much like the existing ingredient handlers, except that it receives a `section` tree and a `logger` object, instead of the complete page tree. If the section can't be found (i.e., the H2 is missing, `sectionHandler` returns the appropriate error.

New in this PR: if the inner callback function returns `true` (or some other truthy value, for success), then the wrapper handles returning the correct position H2 node. If the callback function returns `false` (or some other falsy value), then the wrapper returns the correct null value.

Additionally, in the course of refactoring, I spotted a suspected bug, confirmed it, and fixed it.

## Interesting commits 

Most of the changes are rather mechanical: wrapping the existing handler functions and changing return values to the right type. Occasionally, I had to substitute an error message's position node with one from the section context.

Unfortunately, it touches lots of lines, thanks to changes in indentation; the total number of lines of code is reduced, excluding tests.

Here are some highlights:

* d40a21d - adds the position node handling to `sectionHandler`
* 9421959 and d35f879 - the most complex refactor, to `checkLinkList`, class methods, and constructors
* c6f0922 and 9a6f74f - adds to tests for a bug I spotted in the link list checker, and fixes it

### DIFF
--- a/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-browser-compatibility.js
+++ b/scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/data-browser-compatibility.js
@@ -1,4 +1,3 @@
-const { select } = require("hast-util-select");
 const visit = require("unist-util-visit");
 
 const utils = require("./utils.js");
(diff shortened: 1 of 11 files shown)


wbamberg (approved):
This looks great to me @ddbeck , thanks for going through with this change.

PR #507, opened 2020-07-08
Add data.permitted_properties handler

### LINKED ISSUE #498: U - Write ingredient handler for `data.permitted_properties`
From #488. We need to be able to lint the ingredient which appears in the the new recipe.

ddbeck:
This PR adds a handler for the `data.permitted_properties` ingredient. It's very nearly identical to `data.constituent_properties`, so I refactored that code to be a factory for property list handlers. As a consequence, there's not much new here, apart from the tests.

Fixes #498.


scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/index.js
```diff
@@ -38,6 +41,7 @@ const ingredientHandlers = {
   "data.instance_methods?": classMembers.handleDataInstanceMethods,
   "data.instance_properties?": classMembers.handleDataInstanceProperties,
   "data.interactive_example?": handleDataInteractiveExample,
+  "data.permitted_properties": handleDataPermittedProperties,
```
wbamberg, line 44: It's actually `data.permitted_properties?`, which probably means that the factory needs to get and understand an `isOptional` flag.

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/property-lists.js
```diff
@@ -7,15 +7,27 @@ const {
...
+
+function propertyListHandler(details) {
+  return sectionHandler(details.id, (section, logger) => {
+    const expectedIntroTextP = findIntroTextP(section, details.introText);
 
     if (expectedIntroTextP === null) {
       logger.expected(
         section,
-        '"This property is a shorthand for the following CSS properties:" paragraph',
+        `"${details.introText}" paragraph`,
```
wbamberg, line 30: I understand that this is in line with the existing handler, but I think this gives much too wordy an error message, which is less readable than something like "Expected specific intro paragraph" or something.

wbamberg (changes requested):
Thanks for this @ddbeck ! This is also a very nice refactor. But it needs to handle the fact that `data.permitted_properties?` is optional.

This could be more generic than "properties", too: it's really any section which is a UL of `code` links with an intro paragraph.

In fact now I think about it, why don't we use "link lists" for these things? Why are we using a UL with no short description for shorthand properties, and a DL with short descriptions for (say) lists of methods? Is there actually a good reason for this inconsistency? (I know you have put work into this and I'm sorry to think of it only now. But now I have thought of it I feel compelled to share.)

ddbeck:
That's a really helpful review, Will. Thank you! I'll being doing these things shortly:

- [ ] Rename the wrapper to reflect that it goes for any alpha-sorted list with intro text
- [ ] Shorten intro text message and move full-length text to the message's notes
- [ ] Add tests for the ingredient's optionality
- [ ] Implement optionality

Regarding you question about link lists:

> why don't we use "link lists" for these things? Why are we using a UL with no short description for shorthand properties, and a DL with short descriptions for (say) lists of methods? Is there actually a good reason for this inconsistency?

It's a good question and you're right to ask about it. If anything, I'd turn this around: why aren't we using ULs for method lists?

I don't like the idea of using link lists for property lists for a few reasons:

* Reading: names and descriptions seem harder to scan than a bare list (admittedly, this is mostly presentation)
* Scraping: we don't have a stumptown answer for DLs yet, so I'd rather not starting using _more_ definition lists
* Authoring: writing DLs is fussier than writing plain lists
* Implementation: it seems like it would be a lot of extra work to get one-line descriptions of every permitted or constituent property onto the pages where we'll be adding these sections. That feels like the _reward_ of structuring the content this way, not a prerequisite

(To answer my own question, we probably shouldn't change the existing method lists either. It's a bit of a regression if you're expecting that descriptive text and it's also a lot of work to remove that stuff.)

ddbeck:
OK, I've pushed commits fixing the stuff I put in the checklist above.

I had a hard time renaming the module. Really, I should reorganize all the ingredient utils: there's handler factories in `index.js`, `utils.js`, and now `list-sections.js`. I started to fix this, but I think it made the diff pretty noisy, so I'll open a follow-up PR for that.

wbamberg:
> Regarding you question about link lists:
> 
> > why don't we use "link lists" for these things? Why are we using a UL with no short description for shorthand properties, and a DL with short descriptions for (say) lists of methods? Is there actually a good reason for this inconsistency?
> 
> It's a good question and you're right to ask about it. If anything, I'd turn this around: why aren't we using ULs for method lists?
> 
> I don't like the idea of using link lists for property lists for a few reasons:
> 
> * Reading: names and descriptions seem harder to scan than a bare list (admittedly, this is mostly presentation)
> 
> * Scraping: we don't have a stumptown answer for DLs yet, so I'd rather not starting using _more_ definition lists
> 
> * Authoring: writing DLs is fussier than writing plain lists
> 
> * Implementation: it seems like it would be a lot of extra work to get one-line descriptions of every permitted or constituent property onto the pages where we'll be adding these sections. That feels like the _reward_ of structuring the content this way, not a prerequisite
> 
> 
> (To answer my own question, we probably shouldn't change the existing method lists either. It's a bit of a regression if you're expecting that descriptive text and it's also a lot of work to remove that stuff.)

I think it's definitely arguable that UL is better than DL. However:

* three of your four reasons seem to be appealing to writers, not readers. And your "a bit of a regression" comment implies that in some ways at least DL is better for readers. And we should prioritize readers, surely.

* just as a practical matter, DL is the incumbent. Hundreds of pages use them for lists of links. To change it we'd need to convince everyone that ULs are better, and even then it would be a ton of work to turn existing DLs into ULs. Conversely, it would be uncontroversial and fairly quick to write DLs for constituent properties (about 35 pages) and permitted properties (probably <10 pages).

So there are basically three options:
1. be consistent and use DLs
2. be consistent and use ULs (which I think we can exclude just on the amount of work it is)
3. be inconsistent

I'm OK with 3 as long as we're clear about it.

Of course in stumptown, this is treated as a presentational issue, not a content issue at all. The content is a list of references to pages, and the renderer gets to decide whether to make them a UL, or a DL, or a table, or.. And you can change it all in one place, and decide which you like the look of better, and change it back if you change your mind. But, well,  ¯\_(ツ)_/¯.

wbamberg:
BTW though, I love the way this exercise makes us ask these kinds of questions about high-level consistency across our docs!

ddbeck:
> BTW though, I love the way this exercise makes us ask these kinds of questions about high-level consistency across our docs!

Me too. This is enjoyable, interesting, important work.

> three of your four reasons seem to be appealing to writers, not readers. And your "a bit of a regression" comment implies that in some ways at least DL is better for readers. And we should prioritize readers, surely.

First, maybe I should've said _perceived_ regression. I was thinking of this along the lines of the spec tables: we ultimately got a complaint about the loss of historic spec links. The spec tables contain less information now, that's plainly a loss to people who relied on them. But I think those tables are better for our audience in general, despite that loss to some readers (and I was, for the record, skeptical of that decision—I would've chosen to keep the historic links and I am happy to report that I would've been wrong).

You're also right that I was thinking mostly for the benefit of writers, but I'd argue that maintenance costs are eventually paid by the reader.

Considering CSS alone, if we go with DLs for property lists right now, it means you're going to soon end up with several slightly inconsistent short descriptions for every property mentioned in a DL property list (one for every property list it appears in, plus the page for that property). It's a different story once we're in Yari (and again when we get into structured markdown), but adopting this _now_ is adopting at its point of greatest maintenance cost.

I'd wager that the cost to the reader of clicking through to a specific property is less than the benefit they might gain from us not taking on that maintenance burden (though I'll concede it's a high-risk wager—there are lots of readers and lots of clicks). To me, that rules out option 1.

Having said all that, I've convinced myself that option 2 (ULs only) is preferable: I _would_ like to get rid of the definition lists. I'd be very tempted to do that in bulk, once we're in Git. I'm not even sure it's that much work, really. You could probably turn the linter code against this problem pretty easily, but doing so without the net of version control is too scary to me. It's a bit like removing the `jsxref` macro.

So, we're left with option 3: be inconsistent. I feel good about this. We'd be inconsistent for the benefit of avoiding other inconsistencies. 😄

wbamberg:
> Having said all that, I've convinced myself that option 2 (ULs only) is preferable: I would like to get rid of the definition lists. I'd be very tempted to do that in bulk, once we're in Git. I'm not even sure it's that much work, really. You could probably turn the linter code against this problem pretty easily, but doing so without the net of version control is too scary to me. It's a bit like removing the jsxref macro.

👍 

> So, we're left with option 3: be inconsistent. I feel good about this. We'd be inconsistent for the benefit of avoiding other inconsistencies. 😄

👍

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/list-sections.js
```diff
@@ -0,0 +1,186 @@
...
+          `Expected text: ${details.introText}`
+        );
+        return null;
+      }
+
+      const expectedUl = findNextUl(expectedIntroTextP, section);
+      if (expectedUl === null) {
+        logger.expected(
+          expectedIntroTextP,
+          "property list UL after intro text",
```
wbamberg, line 42:
```suggestion
          "Property list UL after intro text",
```

scripts/scraper-ng/rules/html-require-recipe-ingredients/ingredient-handlers/list-sections.js
```diff
@@ -0,0 +1,186 @@
...
+        );
+        return null;
+      }
+
+      const lis = findLis(expectedUl);
+
+      if (lis.length < details.minimumListItems) {
+        logger.expected(
+          expectedUl,
+          "two or more LIs in property list",
```
wbamberg, line 53:
```suggestion
          "Two or more LIs in property list",
```

wbamberg (changes requested):
This looks great @ddbeck , thanks for the quick turnaround. I had a couple of nitpicky comments about case: we seem to prefer a capital letter at the start of the error message. Otherwise this looks and seems to work great.

ddbeck:
I'm not sure capitalizing `loger.expected(`) calls is right. The template for that is: `Expected ${name} for ${ingredient}`, so I think lower case works there a bit better. `logger.fail()` just takes a string, where we should capitalize, or maybe decide no capitalization everywhere is a better call. I don't know.

But I do know we already talked about making this a generic wrapper for any section+intro+UL+code, so I'm pushing a change that rewrites the messages to not mention _property_ lists specifically.

wbamberg (approved):
Oh, sorry, I misread that.

Then this looks good to go! Thanks again Daniel!

PR #508, opened 2020-07-08
Add recipe for CSS keywords

wbamberg:
As discussed in https://github.com/mdn/sprints/issues/3171#issuecomment-655197948, this adds a new recipe for CSS keywords. I've tagged the keyword pages:

https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/inherit
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/initial
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/revert
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/unset

...with "Keyword", and with this PR and that change we now see only two "Recipe missing" messages.

### DIFF
--- a/recipes/css-keyword.yaml
+++ b/recipes/css-keyword.yaml
@@ -1,0 +1,12 @@
+data:
+  - title
+  - short_title?
+  - mdn_url
+body:
+  - prose.short_description
+  - prose.description?
+  - prose.*
+  - data.examples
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
Thank you, Will! 🎉

